PR #3, opened 2020-07-15
Create CI pipeline

ddugovic:
GitHub offers a free CI pipeline for open source projects, and I find in my own projects e.g. https://github.com/ddugovic/Stockfish that having pipelines helps onboard curious developers (and helps w/ root cause analysis):
https://github.com/features/actions

Thanks for your excellent lectures published on YouTube!

(As for why I submit this PR... this decade I'd like to see CI/CD and TDD more widely advocated for across the entire tech sector, and it's never before been easier to get started and experiment!)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,0 +1,23 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: configure
+      run: ./configure
+    - name: make
+      run: make
+    - name: make check
+      run: make check
+    - name: make distcheck
+      run: make distcheck
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -15,19 +15,3 @@ jobs:
       run: cmake --build build
     - name: test
       run: cd build && ctest
-
-  build-windows:
-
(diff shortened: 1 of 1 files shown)


mattgodbolt:
Oh that's great: thanks! I've been using travis in other projects, but this is a great opportunity to lean how GitHub's own CI system works! Cheers!

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  build-ubuntu:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v1
+    - name: configure
+      run: mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" ..
```
mattgodbolt, line 13: we already have `-Werror -Wall -Wextra` specified in the CMake itself: will this override them or be in addition to?

Also with a more modern `cmake` installation one can do `cmake -S . -B build ...` too . I think ubuntu latest probalby has that. In fairness I had to undo those changes from the `Makefile`as the system I was testing on had an old cmake...

ddugovic: Good question and I fear it may be the former (override) which was not my intent when copying these steps from a comment in https://github.com/actions/starter-workflows/pull/47 .

Oh, very nice! CMake continues to impress me... for years (for open source hobby projects) I stubbornly chose Autotools for imagined portability advantages (with other people running my chess AI or GUI code) which seem to be a red herring (now if another developer wants to try supporting another platform, they can PR a test and maybe even a patch!).

Cheers!

PR #12, opened 2020-07-22
First pass at a big reformat and the make target

mattgodbolt:
Give this a look and see what you think. We can turn off formatting with some magic markers.

If you don't like the look of a particular thing, let me know. I'll update the clang-format settings and then reset the `src` back to how it was in `main`, then re- `make reformat-code`


mattgodbolt:
Based on conversations we had I think this is ready to go. Will rebase it on main and then we can deal with the fallout as time progresses!

PR #13, opened 2020-07-23
Conan, C++ and tests

mattgodbolt:
This PR introduces `conan` (a C++ package manager), `Catch2` (a C++ unit test framework, brought in via `conan`), and makes a first stab at writing some tests for a piece of functionality: the `buffer_*` functions.

The intention is to write some tests, consider reimplementing the complex logic in the `struct buffer` as a C++ class (with a C wrapper for compatibility), check the tests still pass, then see if we can switch out the C code for C++ "fully". The last step may be trickiest, but hoping the buffer.c can be replaced with about ten lines of actual C++ code.


mattgodbolt:
Most of the change is conan/cmake/makefile guffance. It should still Just Work™ for your workflow (and IDE workflows). There was a bit of nasty extraction to get `main` out of `comm.c` and then the test is src/test/buffer_test.cpp

src/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 add_subdirectory(doorman)
 add_subdirectory(chat)
-add_executable(xania
+add_subdirectory(test)
+add_library(xania_lib STATIC
```
snellers, line 4: Aha, a static library at last :)

snellers: Megapedantry but I'm not bothered:  libxania_lib.a == double lib :) It could perhaps just be libxania.a

src/comm.c
```diff
@@ -8,39 +8,26 @@
...
- *    Game_loop ---> Read_from_descriptor ---> Read
- *    Game_loop ---> Read_from_buffer
- *
- * The data flow for output is:
- *    Game_loop ---> Process_Output ---> Write_to_descriptor -> Write
- *
- * The OS-dependent functions are Read_from_descriptor and Write_to_descriptor.
- * -- Furey  26 Jan 1993
+ * Originally contained a ton of OS-specific stuff.
+ * We only support linux these days...so we stopped pretending...
```
snellers, line 12: RIP my dreams of running the mud on an ebay'd SGI Onyx RealityEngine

src/main.c
```diff
@@ -0,0 +1,118 @@
...
+#include <signal.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/time.h>
+#include <unistd.h>
+
+/* SIGTRAP on/off */
+bool debug = FALSE;
+/* check the mobs/objects */
+bool printinfo = FALSE;
```
snellers, line 14: Ooooh...I'd forgotten about this. It calls report_object(),  report_mobile().  This could be very useful for balancing efforts.

src/test/buffer_test.cpp
```diff
@@ -0,0 +1,96 @@
+#include "buffer.h"
+
+#include <catch2/catch.hpp>
+#include <string_view>
+
+namespace {
+
+// This could become the beginnings of a C++ wrapper over a C implementation...
+// And then eventually a reimplementation in C++ (which maybe a C shim if needed).
+class Buffer {
```
snellers, line 10: OK so this is just a "demo" class to illustrate testability.

snellers: Ah I see, it wraps the older buffer code.

src/test/buffer_test.cpp
```diff
@@ -0,0 +1,96 @@
...
+            b.removeline();
+            CHECK(b.sv() == "This is a whole line\nA"); // yes, really
+        }
+    }
+    SECTION("Should handle printf-style formats") {
+        b.addline_fmt("What an %s way to print the number %d\n\r", "anachronistic", 12);
+        CHECK(b.sv() == "What an anachronistic way to print the number 12\n\r");
+    }
+    SECTION("Should handle really big documents and expand") {
+        constexpr auto NumLines = 10'000u;
```
snellers, line 80: Is this a naming convention you tend to use for constexprs -- UpperCamel style?  Seems fine, I've seen it used it used like that in Java too (it's cosmetic, but I do like it more than UPPER_SNAKE most of the time).

src/CMakeLists.txt
mattgodbolt: Sadly we already have a `xania` target in CMake-land :( (the executable). So we either have a "xania" binary or a "xania" lib, but not both

src/comm.c
mattgodbolt: Sorry mang :P

src/main.c
mattgodbolt: So much crupht but yeah

src/test/buffer_test.cpp
mattgodbolt: Totally. It's there to show how it could be, while using the exact same code.

src/test/buffer_test.cpp
mattgodbolt: Yeah, `SHOUT_CASE` is for macros only these days (imnsho)

src/CMakeLists.txt
horenmar: You can modify the binary name to fix this. See e.g. 

https://github.com/master-keying/minisat/blob/master/CMakeLists.txt#L52-L57

mattgodbolt: Oh thanks @horenmar: TIL :D

PR #18, opened 2020-07-23
Check format

mattgodbolt:
A failed format doesn't preclude anything else but it will show up in the PR. So we can see if an external PR meets the requirements.


mattgodbolt:
If we want to get super fancy later we can make "code format" and a bunch of other things required to pass.

snellers:
Cool nice

PR #37, opened 2020-07-24
Annotate all printf()ish functions and fix up warnings.

### LINKED ISSUE #26: Mark all variadic printf functions as "printf-like"
We have a bunch of functions that act like `printf()`. We should annotate them with `__attribute__((format))` ( https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html ) so we get compile-time checking.

That will ease us when we start moving from char* to `std::string`

mattgodbolt:
Closes #26


mattgodbolt:
PR only so I can make sure this works on all the compilers.

PR #49, opened 2020-07-25
Make 'install' the default make target so that users must run 'make s…

### LINKED ISSUE #41: 'make' builds and runs which isn't always what you want. Ideally it should only run if you run 'make run' or 'make start'

snellers:
…tart' to start xania and doorman. Update README to reflect this along with some steps about gdb. Resolves #41

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: default
-default: start
+default: install
 
 .PHONY: help
 help: # with thanks to Ben Rady
--- a/README.md
+++ b/README.md
@@ -20,8 +20,9 @@ downloaded and managed by the master `Makefile`.
 To build and run, type `make`:
 
 ```bash
-$ make
+$ make start
 ... lots of configuration nonsense here, downloads, cmake, and then compilation ...
+
 Starting Xania on port 9000}
 (cd src && ./mudmgr -s 9000)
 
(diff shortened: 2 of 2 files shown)


snellers:
fyi @philjessies

Makefile
```diff
@@ -1,5 +1,5 @@
 .PHONY: default
-default: start
```
snellers, line 2: You ok with this?

mattgodbolt: 100%. There's a `make` way of setting the default target too (but I forgot about it when I wrote the makefile..) Purely FYI (and it's make 3.81 and above only):
```
.DEFAULT_GOAL := install
```

etc :)

README.md
```diff
@@ -22,6 +22,13 @@ To build and run, type `make`:
 ```bash
 $ make
 ... lots of configuration nonsense here, downloads, cmake, and then compilation ...
+```
+
+Then to start it up:
+
+```bash
+$ make start
+
```
mattgodbolt, line 31: Maybe we could just change to `make start` on line 23 as that will build and start?

README.md
```diff
@@ -74,8 +81,31 @@ will pick up on this and just use it. Command-line users can run `make reformat-
...
 There are some very early tests. To run the tests, use `make test`, or run them from your IDE.
 
+### Debugging Xania in GDB
+
+There's more than one way to do this, and one way is:  after building Xania you can run 
+it like so using a sub-shell in the `area` directory (use of `gdb` is optional):
+
+```bash
+$ (cd area && gdb ../install/bin/xania)
```
mattgodbolt, line 94: side note, I filed #53 to make this easier)

Issue #46, opened 2020-07-25
Security "audit"


mattgodbolt:
A number of areas of code are the worst kind of C, and likely suffer from buffer overruns (I have an email containing some issues, obviously not putting here).

Thoughts:

* TESTS!
* Fuzzing doorman and the Xania-side client input handling
* Running with the address sanitizer on (why not? machine's fast enough!)

PR #54, opened 2020-07-28
Part five: Doorman rewrite

mattgodbolt:
Now ready, based off #83 now to hopefully show the actual differences.

Previous TODOs:

* [x] Tests
* [x] Fixing the god-awful input parsing and telnet option handling.
* [x] Run some fuzzing/bad behaving clients over it
* [x] Coding convention renames
* [x] Channel to no longer be a static array (and probably use non-repeating channel IDs to get around some race conditions I've noted). Channels written so construction initialises everything and all the horrid pointers-instead-of-reference and magic numbers in `newConnection` can go. CC @philjessies for the comment he made about the crimes committed in `newConnection`
* [x] Replace log_out (with maybe [spdlog](https://github.com/gabime/spdlog) or just `fmt::print(stderr...)`)
* [x] Replace `char buffer[4096]` and similar with some max-sized payload constant, shared between MUD and doorman. See @snellers comment about `MaxIncoming`
* [x] grep for `TODO`, e.g. the `writev` stuff.


src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+    fd_.close();
+    host_lookup_fd_.close();
+    if (host_lookup_pid_) {
+        if (::kill(host_lookup_pid_, 9) < 0)
+            log_out("Couldn't kill child process (ident process has already exited)");
+        host_lookup_pid_ = 0;
+    }
+}
+
+void Channel::newConnection(Fd fd, sockaddr_in address) {
```
snellers, line 98: nit: I notice some methods are snake case, others are camel, you were leaning towards snake right?

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+    }
+    send_to_client(".", 1);
+}
+
+void Channel::close() {
+    fd_.close();
+    host_lookup_fd_.close();
+    if (host_lookup_pid_) {
+        if (::kill(host_lookup_pid_, 9) < 0)
+            log_out("Couldn't kill child process (ident process has already exited)");
```
snellers, line 93: s/ident/dns?

src/doorman/Channel.cpp
snellers: Interesting,  Fd is being passed by value? And I see it uses std::move() to do so.

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+    case 0: {
+        // Child process - go and do the lookup
+        log_out("ID: doing lookup");
+        async_lookup_process(address, std::move(child_fd));
+        log_out("ID: exiting");
+        exit(0);
+    } break;
+    case -1:
+        // Error - unable to fork()
+        log_out("Unable to fork() an IdentD lookup process");
```
snellers, line 147: I think  all the identd strings can be taken out (now or later)

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+        break;
+    default:
+        host_lookup_fd_ = std::move(parent_fd);
+        host_lookup_pid_ = forkRet;
+        log_out("Forked hostname process has PID %d on %d", forkRet, host_lookup_fd_.number());
+        break;
+    }
+}
+
+bool Channel::incomingData(gsl::span<const byte> incoming_data) {
```
snellers, line 157: aha,  the modern way to do char* foo, size_t n  :)

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+        } else {
+            ptr++;
+            nRealBytes++;
+            nLeft--;
+        }
+    }
+    /* Now to update the buffer stats */
+    buffer_.resize(nRealBytes);
+
+    /* Second pass - look for whole lines of text */
```
snellers, line 275: Yeah surely there's a nice modern way to do a string split on newlines :)

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+        host_lookup_pid_ = 0;
+        return;
+    }
+    // Log the source IP but mask it for privacy.
+    log_out("[%d] %s", id_, get_masked_hostname(hostname_).c_str());
+    sendInfoPacket();
+}
+
+void Channel::sendInfoPacket() const {
+    char buffer[4096];
```
snellers, line 332: inconsistent buffer size here compared to MaxIncoming (related to size comment I made earlier today too but we can tackle those separately)

src/doorman/Channel.hpp
```diff
@@ -0,0 +1,83 @@
...
+    void incomingHostnameInfo();
+    void sendTelopts() const;
+
+    void on_auth(std::string_view name) { auth_char_name_ = name; }
+    [[nodiscard]] const std::string &authed_name() const { return auth_char_name_; }
+    void on_reconnect_attempt();
+    void mark_disconnected() noexcept { connected_ = false; }
+    void send_to_client(gsl::span<const char> span) const { send_to_client(span.data(), span.size_bytes()); }
+    void send_to_client(gsl::span<const byte> span) const { send_to_client(span.data(), span.size_bytes()); }
+    void send_to_client(const void *data, size_t length) const {
```
snellers, line 52: could this one be private?

src/doorman/Channel.hpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    // Intention here is to prevent direct access to the fd.
+    [[nodiscard]] bool has_fd(int fd) const { return fd == fd_.number(); }
+    [[nodiscard]] bool is_closed() const { return !fd_.is_open(); }
+
+    void set_echo(bool echo);
+    void close();
+
+    [[nodiscard]] int set_fds(fd_set &input_fds, fd_set &exception_fds) noexcept {
```
snellers, line 64: I note a mix of implementation in the header vs the cpp file. set_dfs() isn't a one-liner.  Do you use a particular rule for that? e.g. simple pure getters can go in the header, or if it's a template class they all have to.

src/doorman/Xania.cpp
```diff
@@ -0,0 +1,176 @@
...
+namespace {
+void *iovec_cast(const void *thing) { return const_cast<void *>(thing); }
+}
+
+void Xania::SendToMUD(const Packet &p, const void *payload) {
+    if (!connected())
+        return;
+    const iovec vec[2] = {{iovec_cast(&p), sizeof(Packet)}, {iovec_cast(payload), p.nExtra}};
+    // TODO writev? rephrase this whole thing?
+    if (::writev(fd_.number(), vec, 2) != static_cast<ssize_t>(sizeof(Packet) + p.nExtra)) {
```
snellers, line 28: Interesting use of writev, we didn't have that before.

src/doorman/doorman.cpp
```diff
@@ -0,0 +1,367 @@
...
+                mud_.process_mud_message();
+            } else {
+                /* It's a message from a user or ident - dispatch it */
+                if (auto it = std::find_if(begin(channels_), end(channels_),
+                                           [&](const Channel &chan) { return chan.is_hostname_fd(i); });
+                    it != end(channels_)) {
+                    it->incomingHostnameInfo();
+                } else {
+                    int nBytes;
+                    byte buf[256];
```
snellers, line 273: We should perhaps have a single constant in a header for this

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+}
+
+void Channel::newConnection(Fd fd, sockaddr_in address) {
+    // One day I will be a constructor...
+    this->fd_ = std::move(fd);
+    hostname_ = inet_ntoa(address.sin_addr);
+    port_ = ntohs(address.sin_port);
+    netaddr_ = ntohl(address.sin_addr.s_addr);
+    buffer_.clear();
+    width_ = 80;
```
philjessies, line 105: Would it not make sense to have 'int width_  = 80;' in the .h file, rather than default-initing it there, and then overwriting it here?

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,393 @@
...
+            /* Send it to the MUD */
+            mud_->on_client_message(
+                *this, std::string_view(reinterpret_cast<const char *>(buffer_.data()), ptr - buffer_.data()));
+            /* Check for \n\r or \r\n */
+            if (nLeft > 1 && (*(ptr + 1) == '\r' || *(ptr + 1) == '\n') && *(ptr + 1) != c) {
+                memmove(buffer_.data(), ptr + 2, nLeft - 2);
+                nLeft--;
+            } else if (nLeft) {
+                memmove(buffer_.data(), ptr + 1, nLeft - 1);
+            } // else nLeft is zero, so we might as well avoid calling memmove(x,y,
```
philjessies, line 295: Not sure I understand this comment. If nLeft is 1, we're going to call memmove(_, _, 0); above anyway; AFAICT what we save with the 'if (nLeft)' check is calling memmove with -1 as the number of bytes. Right?

src/doorman/Channel.cpp
mattgodbolt: > nit: I notice some methods are snake case, others are camel, you were leaning towards snake right?

Absolutely: I plan on a big rename as part of the "Coding convention renames"

>Interesting, Fd is being passed by value? And I see it uses std::move() to do so.

Yup! `Fd` is a "move only" type by dint of having its copy constructor (`Fd(const Fd&)`) and copy assign (`Fd &operator=(const Fd&)`) `delete`d. That means (this is non-normative, but mostly correctly) that to have an Fd you have to "own" it, either by having created it yourself, or having `std::move()`d it. That way we know with certainty that if you have an `Fd` you own it. If you let it fall out of scope, it will close. If you need to hand it to another function or object, you have to relinquish it and give it to that object (unless you use a temporary "lending" reference `const Fd &`, but I tried not to do that here).

src/doorman/Channel.cpp
mattgodbolt: Done! Thanks - I also looked for others and replaced them.

src/doorman/Channel.cpp
mattgodbolt: Agreed; as part of the logging changes I'd like to make I think `log_out` will die and then we'll have `debug`, `info`, `warn` levels etc so it's not quite so noisy when we don't want it to be. I'll have a rethink then.

src/doorman/Channel.cpp
mattgodbolt: > aha, the modern way to do char* foo, size_t n :)

In this case `const byte *, size` yeah. 

Aside: There's a language-level thing for strings: `string_view` is `const char*, size_t` and I use it in a bunch of places. It can be automatically converted-to from a `const char *` and a `std::string` so it's a good way to pass a non-owning reference to a string. BUT it's not NUL terminated, so only useful if you don't need that.

src/doorman/Channel.cpp
mattgodbolt: You'd think. But yeah :) We can do a lot better!

src/doorman/Channel.cpp
mattgodbolt: Right. I think a `Packet::MaxPayloadSize` on both sides is probably needed. Will add a TODO in the code and in the PR notes above to cover this.

src/doorman/Channel.hpp
mattgodbolt: It can't but with a trivial tweak to a caller it can! Thanks: fixed!

src/doorman/Channel.hpp
mattgodbolt: MOSTLY if it's trivial I put it in the header. This is an example of something that grew and I lazily didn't move it. With modern compilers, they can inline things that aren't even in the header, and I tend to err on putting more things in the cpp file I'll move this one as it's out of place.

There's a big movement in the C++ world to put _everything_ in headers. I don't subscribe to this viewpoint. I'm a grumpy chap about it:)

src/doorman/Xania.cpp
mattgodbolt: Right, the checking of return values for multiple write() calls was annoying me, so early on in the conversion (before I made it `throw`, something I'm still considering changing back), it's nicer to have one call to check :)

`writev` is great though for exactly this stuff. I think there's a good c++-ey solution to wrapping writev() nicely, hence the `TODO`

mattgodbolt: (have updated the list of things to do in the PR description accordingly)

src/doorman/doorman.cpp
mattgodbolt: Agreed; will use an appropriately-calculated amount so as not to overflow the `Channel::MaxPayload` when I put it in (noted in the PR description).

src/doorman/Channel.cpp
mattgodbolt: Agreed. The way the code is phrased the Channels are statically allocated and then reused over and over again instead of being destructed and re-constructed. That's a holdover from the way Channel IDs work and the way the code was written, and one I plan on changing. Ideally the Channel would be constructed when it has a connection come in, that way all these variables are initialsed there. I hate the fact we overwrite them here. For the constants, having them in the header file (presumably you mean member initialisation here? (`int width_{80};` etc) makes sense then.

mattgodbolt: Noted in the PR TODOs

src/doorman/Channel.cpp
mattgodbolt: The comment is wrong, and the code is heinous. In fact my IDE warns me that nLeft CANNOT be zero at this point. But it's left exactly as it was before for now... I'm not touching any of the parsing code until I have tests for it all. It hurts too much :)

mattgodbolt:
Just ran a couple of hacky overflow attempts (which show up some deficiencies in the MUD source code (CC @apmorton)); doorman's (new) code at least looked alright, with giant messages, random messages, and a few other weird strings. All seems OK, at least no different from the issues we need to solve in #46.

mattgodbolt:
Last TODO done. Gonna review this myself and see if I can't get it broken into a few smaller (ish) PRs.

src/doorman/Doorman.cpp
```diff
@@ -0,0 +1,126 @@
...
+        auto newFd = listenSock_.accept(reinterpret_cast<sockaddr *>(&incoming), &len);
+
+        if (channels_.size() == MaxChannels) {
+            newFd.write("Xania is out of channels!\n\rTry again soon\n\r"sv);
+            log_.warn("Rejected connection - out of channels");
+            return;
+        }
+
+        auto id = id_allocator_.reserve();
+        auto insert_rec = channels_.try_emplace(id->id(), *this, mud_, id, std::move(newFd), incoming);
```
snellers, line 101: One style comment:  I was trying to track down where we were instantiating Channels. And by hiding it in a try_emplace() it makes it a bit tricky to do that :)  Probably the same for any other collections we're writing to with an "inlined" construction rather than a local var.

src/comm.c
```diff
@@ -345,23 +360,32 @@ void game_loop_unix(int control) {
         if (doormanDesc && FD_ISSET(doormanDesc, &in_set)) {
             do {
                 Packet p;
-                char buffer[4096 * 4];
+                char buffer[UMAX(PACKET_MAX_PAYLOAD_SIZE, 4096 * 4)];
```
snellers, line 363: Bit concerned about the size of this. It's always going to be 16384. But it's then writing into d->inbuf in read_from_descriptor(), which is 4*MAX_INPUT_LENGTH.    Plus, doorman itself in Xania.send_to_mud won't send anything longer than PACKET_MAX_PAYLOAD_SIZE.  I'm wondering if PACKET_MAX_PAYLOAD_SIZE can be set to be equal to MAX_INPUT_LENGTH?

snellers: (granted I see there's a size check on line 373, but it's still larger than inbuf

src/doorman/Channel.cpp
```diff
@@ -0,0 +1,199 @@
...
+    FD_SET(fd_.number(), &exception_fds);
+    FD_SET(fd_.number(), &input_fds);
+    return fd_.number();
+}
+
+void Channel::on_data_available() {
+    // This constant doesn't control how large the data sent to the MUD is; it's more a "fairness"
+    // indicator: per poll() we won't read more than from each socket. This prevents one chatty client
+    // from starving out the others. If a client gets particularly far ahead, then their TCP window
+    // from starving out the others. If a client gets particularly far ahead, then their TCP window
```
snellers, line 155: copied line here?

src/doorman/Doorman.cpp
mattgodbolt: Agreed: sometimes we're stuck with this as one can't move the object one creates: that's the case here. We can't make a temporary Channel and then put it in the map, we must construct it "in place" inside the map. That's a side effect of the fact the channels' address must remain consistent as it is passed as a reference around (particularly as a param to its internal telnet handler), else we could "move" it. `try_emplace` was added in C++17 specifically for this case.

An alternative is to use a `unique_ptr` to a Channel and store that in the map: then we only ever need to create it once in a known location, which is an alternative. Slightly higher overhead (a pointer chase) and minor cognitive load (`->` instead of `.`). Given this caused a bit of confusion, I'll rephrase that way, as we don't expose the pointer-ness or value-ness of the Channel elsewhere (yay encapsulation).

src/comm.c
mattgodbolt: So: I'm trying not to fix things in the mud at this point. the max payload size stuff is a protocol thing between doorman and xania: I don't think we should rely on the mud not overflowing based on it. We should fix that as part of #46 instead of hackily getting buffer sizes to have some kind of relationship in the code I think.

I made the minimum possible change here to at least have some agreed maximum size between xania and doorman without thinking too much: any suggestions what I should do here as part of this change?

src/doorman/Channel.cpp
mattgodbolt: Ooh cripes

src/comm.c
mattgodbolt: Made some changes per discussion in immchat; please take a look and lmk if you think this is OK as is for now, relying mostly on #46 to handle the input-from-doorman overflowing xania buffers side.

mattgodbolt:
Thanks all: will merge and be damned after I get back from being out. I'll then babysit an upgrade and make sure all works out ok!

PR #80, opened 2020-08-01
Part one of doorman PR (see #54)

mattgodbolt:
* Introduce a file descriptor C++ wrapper.
* Add a header to define byte

### DIFF
--- a/src/doorman/Fd.cpp
+++ b/src/doorman/Fd.cpp
@@ -1,0 +1,86 @@
+#include "Fd.hpp"
+
+#include <fmt/format.h>
+
+using namespace fmt::literals;
+
+void Fd::write(const void *data, size_t length) const {
+    if (!is_open())
+        throw std::runtime_error("Write called on invalid file descriptor");
+    auto num = ::write(fd_, data, length);
+    if (num < 0)
+        throw fmt::system_error(errno, "Unable to write to {}", fd_);
+    if (static_cast<size_t>(num) != length)
+        throw std::runtime_error("Truncated write to file descriptor {}"_format(fd_));
+}
+
+void Fd::writev(gsl::span<const iovec> io_vecs) const {
+    if (!is_open())
+        throw std::runtime_error("Write called on invalid file descriptor");
+    auto num = ::writev(fd_, io_vecs.data(), io_vecs.size());
+    if (num < 0)
+        throw fmt::system_error(errno, "Unable to write to {}", fd_);
+
+    size_t length = 0;
+    for (auto &v : io_vecs)
+        length += v.iov_len;
+
+    if (static_cast<size_t>(num) != length)
+        throw std::runtime_error("Truncated write to file descriptor {}"_format(fd_));
+}
+
+Fd Fd::accept(sockaddr *address, socklen_t *socklen) const {
+    auto accepted_fd = ::accept(fd_, address, socklen);
+    if (accepted_fd < 0)
+        throw fmt::system_error(errno, "Unable to accept from file descriptor {}", fd_);
+    return Fd(accepted_fd);
+}
+
+Fd Fd::socket(int domain, int type, int protocol) {
+    int socket_fd = ::socket(domain, type, protocol);
+    if (socket_fd < 0)
+        throw fmt::system_error(errno, "Unable to create a socket");
+    return Fd(socket_fd);
+}
+
+const Fd &Fd::setsockopt(int level, int optname, const void *optval, socklen_t optlen) const {
+    if (!is_open())
+        throw std::runtime_error("Setsockopt called on invalid file descriptor");
+
+    if (::setsockopt(fd_, level, optname, optval, optlen) < 0)
+        throw fmt::system_error(errno, "Unable to set socket option {}:{}", level, optname);
+    return *this;
+}
+
+const Fd &Fd::bind(const sockaddr *address, socklen_t socklen) const {
+    if (::bind(fd_, address, socklen) < 0)
+        throw fmt::system_error(errno, "Unable to bind to address");
+    return *this;
+}
+
+const Fd &Fd::listen(int backlog) const {
+    if (::listen(fd_, backlog) < 0)
+        throw fmt::system_error(errno, "Unable to listen");
+    return *this;
+}
+
+const Fd &Fd::connect(const sockaddr *address, socklen_t socklen) const {
+    if (::connect(fd_, address, socklen) < 0)
+        throw fmt::system_error(errno, "Unable to connect");
+    return *this;
+}
+
+void Fd::read_all(void *data, size_t length) const {
+    auto num_read = ::read(fd_, data, length);
+    if (num_read < 0)
+        throw fmt::system_error(errno, "Unable to read from {}", fd_);
+    if (static_cast<size_t>(num_read) != length)
+        throw std::runtime_error("Truncated read from file descriptor {}"_format(fd_));
+}
+
+size_t Fd::try_read_some(gsl::span<byte> span) const {
+    ssize_t num_bytes = ::read(fd_, span.data(), span.size());
+    if (num_bytes < 0)
+        throw fmt::system_error(errno, "Unable to read from {}", fd_);
+    return static_cast<size_t>(num_bytes);
+}
--- a/src/doorman/Fd.cpp
+++ b/src/doorman/Fd.cpp
@@ -11,7 +11,7 @@ void Fd::write(const void *data, size_t length) const {
     if (num < 0)
         throw fmt::system_error(errno, "Unable to write to {}", fd_);
     if (static_cast<size_t>(num) != length)
-        throw std::runtime_error("Truncated write to file descriptor {}"_format(fd_));
+        throw std::runtime_error("Truncated write to file descriptor {} ({}/{})"_format(fd_, num, length));
 }
 
 void Fd::writev(gsl::span<const iovec> io_vecs) const {
@@ -26,7 +26,7 @@ void Fd::writev(gsl::span<const iovec> io_vecs) const {
         length += v.iov_len;
 
     if (static_cast<size_t>(num) != length)
-        throw std::runtime_error("Truncated write to file descriptor {}"_format(fd_));
+        throw std::runtime_error("Truncated write to file descriptor {} ({}/{})"_format(fd_, num, length));
 }
 
 Fd Fd::accept(sockaddr *address, socklen_t *socklen) const {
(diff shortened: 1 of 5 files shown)


src/doorman/Fd.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto num = ::writev(fd_, io_vecs.data(), io_vecs.size());
+    if (num < 0)
+        throw fmt::system_error(errno, "Unable to write to {}", fd_);
+
+    size_t length = 0;
+    for (auto &v : io_vecs)
+        length += v.iov_len;
+
+    if (static_cast<size_t>(num) != length)
+        throw std::runtime_error("Truncated write to file descriptor {}"_format(fd_));
```
philjessies, line 29: This "..."_format(fd_) looks weird to me. I'd kind of expect a comma there. Is _format some macro that emits one (or more) things with a comma in front?
Related: why do you use _format(fd_) in some places, and '"... {}", fd_' in others? Should they be consistent?

src/doorman/Fd.cpp
```diff
@@ -0,0 +1,86 @@
...
+Fd Fd::socket(int domain, int type, int protocol) {
+    int socket_fd = ::socket(domain, type, protocol);
+    if (socket_fd < 0)
+        throw fmt::system_error(errno, "Unable to create a socket");
+    return Fd(socket_fd);
+}
+
+const Fd &Fd::setsockopt(int level, int optname, const void *optval, socklen_t optlen) const {
+    if (!is_open())
+        throw std::runtime_error("Setsockopt called on invalid file descriptor");
```
philjessies, line 48: While starting a sentence with a capital is great, 'setsockopt' is a case-sensitive function name. I don't have strong feelings on this, but my vague inclination would be to leave the 's' lower-case.

src/doorman/Fd.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    if (::setsockopt(fd_, level, optname, optval, optlen) < 0)
+        throw fmt::system_error(errno, "Unable to set socket option {}:{}", level, optname);
+    return *this;
+}
+
+const Fd &Fd::bind(const sockaddr *address, socklen_t socklen) const {
+    if (::bind(fd_, address, socklen) < 0)
+        throw fmt::system_error(errno, "Unable to bind to address");
```
philjessies, line 57: Definitely worth reporting the address here.

src/doorman/Fd.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+const Fd &Fd::listen(int backlog) const {
+    if (::listen(fd_, backlog) < 0)
+        throw fmt::system_error(errno, "Unable to listen");
+    return *this;
+}
+
+const Fd &Fd::connect(const sockaddr *address, socklen_t socklen) const {
+    if (::connect(fd_, address, socklen) < 0)
+        throw fmt::system_error(errno, "Unable to connect");
```
philjessies, line 69: I'd report the address that failed here, too.

src/doorman/Fd.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto accepted_fd = ::accept(fd_, address, socklen);
+    if (accepted_fd < 0)
+        throw fmt::system_error(errno, "Unable to accept from file descriptor {}", fd_);
+    return Fd(accepted_fd);
+}
+
+Fd Fd::socket(int domain, int type, int protocol) {
+    int socket_fd = ::socket(domain, type, protocol);
+    if (socket_fd < 0)
+        throw fmt::system_error(errno, "Unable to create a socket");
```
philjessies, line 42: Include domain, type, protocol in the message?

src/doorman/Fd.cpp
```diff
@@ -0,0 +1,86 @@
...
+        throw fmt::system_error(errno, "Unable to connect");
+    return *this;
+}
+
+void Fd::read_all(void *data, size_t length) const {
+    auto num_read = ::read(fd_, data, length);
+    if (num_read < 0)
+        throw fmt::system_error(errno, "Unable to read from {}", fd_);
+    if (static_cast<size_t>(num_read) != length)
+        throw std::runtime_error("Truncated read from file descriptor {}"_format(fd_));
```
philjessies, line 78: Report num_read and length?

src/doorman/Fd.hpp
```diff
@@ -0,0 +1,105 @@
...
+    T read_all() const {
+        T object;
+        read_all(&object, sizeof(object));
+        return object;
+    }
+    template <typename T>
+    void read_all(T &object) const {
+        read_all(&object, sizeof(object));
+    }
+    void read_all(void *data, size_t length) const;
```
philjessies, line 79: Do we need this function? It's very C and dangerous. What about passing in a std::vector<byte>& instead of void*?

src/doorman/Fd.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+    [[nodiscard]] size_t try_read_some(gsl::span<byte> span) const;
+
+    template <typename T>
+    const Fd &setsockopt(int level, int optname, const T &optval) const {
```
philjessies, line 84: Can you please add a comment about what level, optname and optval are? An example or two would be great, and a pointer to where to find the definitions of what optnames are valid maybe.

src/doorman/Fd.hpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+    const Fd &bind(const sockaddr *address, socklen_t socklen) const;
+    template <typename T>
+    const Fd &connect(const T &address) const {
+        static_assert(sizeof(T) >= sizeof(sockaddr));
+        return connect(reinterpret_cast<const sockaddr *>(&address), sizeof(T));
+    }
+    const Fd &connect(const sockaddr *address, socklen_t socklen) const;
+    const Fd &listen(int backlog) const;
+
```
philjessies, line 102: Not clear what 'backlog' means. How many bytes the caller knows are pending?

src/doorman/Fd.hpp
```diff
@@ -0,0 +1,105 @@
...
+    Fd accept(sockaddr *address, socklen_t *socklen) const;
+    static Fd socket(int domain, int type, int protocol);
```
philjessies, line 104: Comment for what the args mean would be great. Examples are always nice too.

src/doorman/Fd.cpp
mattgodbolt: The `_format` suffix is a UDL (user defined literal). I _thought_ I was consistent in my use of formatting but there are two spots where I use the `fmt::format("...")` version (but not in this PR...) Do you have a link to where you  mean? I agree consistency (and will fix the ones in `Channel.cpp` that I found in the appropriate PR).

---

So! Behind the scenes, a UDL is a suffix you can put on a literal to turn it into any object. `fmtlib` (which is being adopted into C++20) makes a format object when you say `"blah"_format` -- basically it's the same as `createFormattingObject("blah")`. The object has an `operator()` which then formats a string with the given params. So:

```cpp
auto x = "Hello {}"_format(world);
```

is equivalent to:

```cpp
auto format_obj = MakeAFormatObj("Hello {}");
auto x = format_obj(world);
```

It looks weird BUT  the compiler can _compile time check_ the parameters: https://godbolt.org/z/PY9qMM

src/doorman/Fd.cpp
mattgodbolt: Yeah: agreed. changed

src/doorman/Fd.cpp
mattgodbolt: It's non-trivial to turn an arbitrary `sockaddr` address into a string without knowing if it's an INADDR address or a unix domain or... That's why I didn't in this case: it's a lot of work for generality. Will make a  TODO to come back to this though.
(If you have tips on how to do the printing though, maybe I've missed something!)

src/doorman/Fd.cpp
mattgodbolt: Done, only as numeric though so useful but not as useful as the name (again no code I know of exists to turn 4 into "AF_INET" but it's better than nothing).

src/doorman/Fd.cpp
mattgodbolt: see below re: address formatting.

src/doorman/Fd.cpp
mattgodbolt: Done (and ditto on the `write` calls)

src/doorman/Fd.hpp
mattgodbolt: This is designed as a thin, C++ safe wrapper over the OS routines: I'm not against documenting but the man page for `setsockopt` is very long. See that, instead

src/doorman/Fd.hpp
mattgodbolt: Again, this is a wrapper over an existing OS routine: if you're using the Fd class, you're using a C++ wrapper over existing OS functionality (and I'm not trying to hide or explain that :))

src/doorman/Fd.hpp
mattgodbolt: see `man listen(2)`

src/doorman/Fd.hpp
mattgodbolt: This function is used by the above function. We can't use a std::vector<> as not all data is being read into a vector: very commonly we're reading into a `Packet` object (using the above). I agree it's dangerous, and I'll see if I can make it at least `private` and only call with the `gsl::span<>` which covers essentially the same base as the `vector` you describe but works for any contiguous array of memory.

In fact, using `gsl::span` is a big improvement, so will take the time to use it more often! Thanks :)

src/doorman/Fd.hpp
```diff
@@ -0,0 +1,119 @@
...
+    void write(gsl::span<const byte> span) const { write(span.data(), span.size_bytes()); }
+    void write(const void *data, size_t length) const;
+    template <typename T>
+    void write(const T &t) const {
+        static_assert(std::is_trivially_copyable_v<T>);
+        write(&t, sizeof(t));
+    }
+    template <typename... Args>
+    void write_many(Args &&... args) const {
+        writev(iovecs_for(std::forward<Args>(args)...));
```
snellers, line 70: Its been a while since I played around with perfect forwarding. And I noticed that noticed that the two places where Xania.cpp currently calls write_many(), it seems to passing a mixture of lvalues and rvalues. I guess this perfect forwarding doesn't care :)

src/doorman/Fd.hpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    template <typename T>
+    T read_all() const {
```
snellers, line 74: It looks like there's one place that calls this version: Xania.cpp:84, to read a Packet. And you would expect the compiler to use return value optimization in that case?

src/doorman/Fd.hpp
mattgodbolt: Right! In honest this falls into "cargo cult" territory: if one accepts a list of template r-value references, one should use `std::forward`. This pattern for varargs "anything" params is one I definitely trot out from memory.

src/doorman/Fd.hpp
mattgodbolt: Absolutely! Named return value optimization (nRVO): https://godbolt.org/z/5YzMYa you'll see that the call to `read_all<BigThing>` on line 11 sets up `rdi` to point on the stack. `rdi` is the first param...the example function takes none! The compiler behind the scenes has rewritten it to be:

```
template <typename T>
[[gnu::noinline]] T read_all(T &a_t_in_the_calling_context) {
  read_all(&a_t_in_the_calling_context, sizeof(a_t_in_the_calling_context));
}
```

Issue #76, opened 2020-07-30
doorman detects ANSI but xania doesn't use it


mattgodbolt:
LIkely not a problem...but we can avoid the "Y/N" prompt iff we remember and use the "ansi" support flag that doorman sends on an info packet (asynchronously). Currently that info's dropped on the floor in comm.c

PR #81, opened 2020-08-01
Part two of doorman PR (see #54)

mattgodbolt:
Builds on #80
* Introduce a standalone, tested telnet protocol handler.


src/doorman/TelnetProtocol.cpp
```diff
@@ -0,0 +1,171 @@
+#include "TelnetProtocol.hpp"
+
+#include <arpa/telnet.h>
+
+#include <algorithm>
+#include <cstring>
+
+static bool supports_ansi(const char *src) {
+    static const char *terms[] = {"xterm",      "xterm-colour", "xterm-color", "ansi",
+                                  "xterm-ansi", "vt100",        "vt102",       "vt220"};
```
philjessies, line 10: Please add "terminator" :-)

src/doorman/TelnetProtocol.cpp
```diff
@@ -0,0 +1,171 @@
...
+        if (body.size() > 1) {
+            std::string term(reinterpret_cast<const char *>(body.data() + 1), body.size() - 1);
+            ansi_ = ::supports_ansi(term.c_str());
+            handler_.on_terminal_type(term, ansi_);
+        }
+        break;
+
+    case TELOPT_NAWS:
+        if (body.size() >= 4) {
+            width_ = (body[0] << 8u) | body[1];
```
philjessies, line 119: Presumably body is a byte*. Will shifting a byte left by 8 preserve the contents because it's already implicitly turned into whatever larger int type width_ is, or do we run the risk of losing the upper 8 bits here? (same for height).

src/doorman/TelnetProtocol.cpp
```diff
@@ -0,0 +1,171 @@
...
+            height_ = (body[2] << 8u) | body[3];
+            handler_.on_terminal_size(width_, height_);
+        }
+        break;
+    default: break;
+    }
+    return body.size() + 3 + 2;
```
philjessies, line 126: While xania is very much a land of magic, I think this particular incantation of '+3+2' deserves a comment (or constant).

src/doorman/TelnetProtocol.hpp
```diff
@@ -0,0 +1,46 @@
...
+    bool echoing_ = true;
+    int width_ = 80;
+    int height_ = 24;
+    bool ansi_ = false;
+    char previous_separator_ = 0;
+
+    void send_com(byte a, byte b);
+    void send_opt(byte a);
+
+    size_t on_subcommand(gsl::span<const byte> command_sequence);
```
philjessies, line 32: A comment about what the return value means would be good. I guess it's the length of the command in bytes, but explicit is nice. (same for on_command)

src/doorman/TelnetProtocol.hpp
```diff
@@ -0,0 +1,46 @@
...
+    size_t on_command(gsl::span<const byte> command_sequence);
+    gsl::span<const byte> interpret_iacs();
+    std::string_view find_whole_lines(std::string_view data);
```
philjessies, line 35: I believe this function's name should be "handle_whole_lines". Find suggests it returns the whole lines (although the return type isn't consistent with that).

Maybe renaming to 'handle_whole_lines' and adding a comment explaining that unhandled trailing text from <data> is returned? Assuming I've understood the function's function correctly.

src/doorman/test/TelnetProtocolTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+    SECTION("Should handle sub options") {
+        SECTION("terminal types") {
+            SECTION("supporting ansi") {
+                REQUIRE_CALL(mock, on_terminal_type("xterm", true));
+                tp.add_data(bytes{IAC, SB, TELOPT_TTYPE, TELQUAL_IS, 'x', 't', 'e', 'r', 'm', IAC, SE});
+                CHECK(tp.supports_ansi());
+            }
+            SECTION("not supporting ansi") {
+                REQUIRE_CALL(mock, on_terminal_type("badger", false));
+                tp.add_data(bytes{IAC, SB, TELOPT_TTYPE, TELQUAL_IS, 'b', 'a', 'd', 'g', 'e', 'r', IAC, SE});
```
philjessies, line 109: [comment] Clearly, the badger terminal is always black and white. Looks spot on to me.

src/doorman/TelnetProtocol.cpp
mattgodbolt: LOL. this list is frankly abominable: it's the one we inherited and really should use `termcap`.  We don't even use it (See #76)

Done though...

src/doorman/TelnetProtocol.cpp
mattgodbolt: `body[0]` is a byte (it comes from the `gsl::span<const byte>` above). Good point, my linter is complaining that this isn't obviously OK, so will change it. It _is_ ok (and even has tests!), so have changed it: thanks for being my conscience here :)

src/doorman/TelnetProtocol.cpp
mattgodbolt: Totally right! Good call. Rephrased in terms of named things (introducing one new name to hopefully make it make more sense!)

src/doorman/TelnetProtocol.hpp
mattgodbolt: Done

src/doorman/TelnetProtocol.hpp
mattgodbolt: Yeah this one's tricksie. I went back and forth a lot on this. Will rename and comment the return value.

src/doorman/TelnetProtocol.cpp
```diff
@@ -0,0 +1,172 @@
...
+            previous_separator_ = 0;
+            continue;
+        }
+
+        auto cr_or_lf = data.find_first_of("\n\r");
+        if (cr_or_lf == std::string_view::npos)
+            break;
+
+        previous_separator_ = data[cr_or_lf];
+        handler_.on_line(data.substr(0, cr_or_lf));
```
snellers, line 78: ok so substr() is  pos, pos+count, so the cr or lf char itself won't get included in the substring

src/doorman/TelnetProtocol.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "Byte.hpp"
+
+#include <gsl/span>
+
+#include <vector>
+
+class TelnetProtocol {
+public:
+    struct Handler {
```
snellers, line 11: I had to poke around for a bit to understand how TelnetProtocol's Handler callback mechanism gets wired up. But I see now:
- Channel implements Handler
- Channel composes Doorman & Xania
- Channel also has an instance of TelnetProtocol and in its construction, its passing a reference to itself (to its "Handler" interface really).

src/doorman/test/TelnetProtocolTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+#include <arpa/telnet.h>
+#include <catch2/catch.hpp>
+#include <catch2/trompeloeil.hpp>
+
+namespace {
+class MockHandler : public TelnetProtocol::Handler {
+public:
+    MAKE_MOCK1(send_bytes, void(gsl::span<const byte>), override);
+    MAKE_MOCK1(on_line, void(std::string_view), override);
```
snellers, line 14: It's not important, but vscode seems to complain about some of these macros, even after I've reconfigured it to look for the headers in ~/.conan/data.

snellers: Aha, I think I resolved that issue.  It was also complaining about std::string_view not being a member of std:: (in this file, but not in some of your others, confusingly).   I went into the C++ extension settings and told it to use the "gnu++17" standard like so, and it looks ok now:

```    "C_Cpp.default.cppStandard": "gnu++17"
```

src/doorman/test/TelnetProtocolTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+using bytes = std::vector<byte>;
+auto wrap(std::string_view sv) {
+    bytes b;
+    for (auto c : sv)
+        b.emplace_back(static_cast<byte>(c));
+    return b;
+}
+auto match_bytes(const bytes &bytes) {
+    return trompeloeil::make_matcher<gsl::span<const byte>>(
```
snellers, line 27: It's interesting how similar this is to Java's Hamcrest Matcher. Same concept, just C++ style.

src/doorman/test/TelnetProtocolTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+}
+
+TEST_CASE("Telnet protocol", "[TelnetProtocol]") {
+    MockHandler mock;
+    TelnetProtocol tp(mock);
+    SECTION("Should start out not supporting ANSI") { CHECK(!tp.supports_ansi()); }
+    SECTION("Should start out empty") { CHECK(tp.buffered_data_size() == 0); }
+    SECTION("Should send appropriate telnet requests and options on request") {
+        REQUIRE_CALL(mock, send_bytes(match_bytes(bytes{IAC, WONT, TELOPT_ECHO})));
```
snellers, line 41: I see it writes the test failures to:  cmake-build-debug/Testing/Temporary/LastTest.log

snellers: I see that Trompeloeil also has a TIMES and IN_SEQUENCE macro.  The Java equivalents would be:
```
verify(mockObject, times(3)).someMethod()
```
and
InOrder:
https://www.javadoc.io/doc/org.mockito/mockito-core/2.6.9/org/mockito/InOrder.html

snellers: And never() is it's FORBID_CALL by the looks of things.

src/doorman/TelnetProtocol.cpp
mattgodbolt: Right :). [substring](https://en.cppreference.com/w/cpp/string/basic_string_view/substr) is `[pos, pos + count)` so this doesn't include the `cr`. I think most (sane!) language substrings work this way (python, Java et al)

src/doorman/test/TelnetProtocolTest.cpp
mattgodbolt: Oh nice! Thanks! We should porbalby put that in the docs. And/or feel free to commit `.vscode` (which I've done in [other projects](https://github.com/compiler-explorer/infra/tree/master/.vscode)).

src/doorman/TelnetProtocol.hpp
mattgodbolt: Sorry it was confusing, it grew a little organically as these things do.

* Channel does indeed implement Handler. It implements it "privately" which means it's not a strict ISA relationship. No user of Channel can "see" that it implements Handler, one can't pass a Channel to something expecting a Handler. BUT Channel itself can pass itself as a handler. The fact that a channel implements an interface is an implementation internal. Another approach would be to have a set of delegates given to TelnetProtocol that all then defer onto various methods in Channel, but this is imo cleaner.
* Channel doesn't strictly compose Doorman and Xania: it has references to them both to be able to remove itself (Doorman interface) and send messages to the mud (Xania interface). In later PRs it might be nicer to literally pass in interfaces to both these things, then we could mock out and test `Channel`s.
* Channel does have a TelnetProtocol and passes itself to the handler interface exactly as you describe.

src/doorman/test/TelnetProtocolTest.cpp
mattgodbolt: Right: tromp forbids all by default, so I don't explicitly call that out here.

mattgodbolt: > I see it writes the test failures to: cmake-build-debug/Testing/Temporary/LastTest.log

If you use `make test`, yeah (`cmake`'s idea of testing is overcomplex IMO).

IDE support usually is better, and if you run the tests directly you get prettier output (esp if they fail).

cmake's testing is test-system agnostic and you can even define tests that are bash scripts etc. It just knows it has to run all the things marked as test, and so to keep things simple it redirects output from each test to a file.

PR #82, opened 2020-08-01
Part three of doorman PR (see #54)

mattgodbolt:
Builds on #81

A general purpose ID allocator. The allocator gives out small integer identifiers, along with a reference-counted reservation that holds them as "in use".

IDs are eagerly recycled once the reference count hits zero.

Useful for Xania and doorman to both have a "vote" for which channels are in-use. Previously we had bugs where there was a race in doorman if someone disconnected while xania was about to send a ton of info while another connection came in. The new connection would be given the channel ID of the recently-closed connection and would thus be delivered all the messages for the old connection.


src/doorman/IdAllocator.hpp
```diff
@@ -0,0 +1,35 @@
...
+        ReservationObj(const ReservationObj &) = delete;
+        ReservationObj(ReservationObj &&) = delete;
+        ReservationObj &operator=(const ReservationObj &) = delete;
+        ReservationObj &operator=(ReservationObj &&) = delete;
+
+        [[nodiscard]] uint32_t id() const noexcept { return id_; }
+    };
+
+public:
+    using Reservation = std::shared_ptr<ReservationObj>;
```
snellers, line 25: This is an interesting convention:  rather than having reserve() state that it returns shared_ptr, it's aliased a "Reservation".  It makes the code shorter so I guess it's more readable. Is that a widespread convention these days?

src/doorman/IdAllocator.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+private:
+    uint32_t next_id_{};
+    std::vector<size_t> released_ids_;
```
snellers, line 29: Should the type of the vector be uint32_t?

snellers:
@mattgodbolt  It's interesting.  I had a thought: whether the shared_ptr technique could be used on the instances of Channel itself, rather than on what appears to be an "integer proxy for a Channel".  I'm sure what's here is fine though.

src/doorman/IdAllocator.hpp
mattgodbolt: It's fairly common. Callers don't need to care that this is a `shared_ptr`, in fact I was going to wrap it in a `Reservation` object that held a shared_ptr, but I was too lazy. Down the line one might wish to do that.

In general, for what it's worth, shared_ptr is considered an anti-pattern. In this case I'm using it very specifically to get the reference counting. Most C++ code gets by without it. It's a common thing you see in code written by folks new to C++ coming from managed languages, though.

src/doorman/IdAllocator.hpp
mattgodbolt: Yes! Amazing spot. I changed from size_t to uin32_t late in the process. I will make an alias to keep the ID the same.

mattgodbolt:
> @mattgodbolt It's interesting. I had a thought: whether the shared_ptr technique could be used on the instances of Channel itself, rather than on what appears to be an "integer proxy for a Channel". I'm sure what's here is fine though.

Right! that was my original design, but it required the Channel to have an "inactive, but known to the MUD" state, which was potentially confusing, and still needed logic to recycle the IDs the channel had. This way the Channel dies as soon as doorman is finished with it (and alongside it, the resources, fds etc, threads/pipes before I tore them out). The trade-off in complexity between either a slightly more complex channel, or an ID mgmt system. I went with the latter mainly as it was a simply-testable component. Maybe not the right choice :)

PR #83, opened 2020-08-01
Part four of doorman PR (see #54)

mattgodbolt:
Builds on #82

Adds a logger, based on spdlog.


src/doorman/Logger.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <string>
+
+#include <spdlog/logger.h>
+
+using Logger = spdlog::logger;
+
+void set_log_level(spdlog::level::level_enum level);
```
philjessies, line 9: Does spdlog::set_level change the log level for all existing loggers, or just new ones created after set_log_level is called? Maybe worth a comment either way.

mattgodbolt: `spdlog::set_level` only sets the log level for the default logger (which we don't really use). my `set_log_level` call here sets it for any created thereafter (ie does the obvious thing you'd probably want). At least any logger created through `Logger.hpp`. I should probably make it loop over existing loggers and change those too. This is more me trying to make spdlog work the way I'd want :grinning:.

mattgodbolt: Ah! Actually the way I'm using `spdlog`, I can't (easily) change existing loggers (I'm trying to avoid the unnecessary `shared_ptr` which seems required to use registered loggers). Unless you think it's a big deal I'll document it and move along.

snellers:
I see there's a bunch of conflicts though^^

mattgodbolt:
> I see there's a bunch of conflicts though^^

Yup will handle. thanks chap

PR #94, opened 2020-08-03
Fix stack overflow seen in apply_prefix() when using an 'at' command.

snellers:


src/interp.c
```diff
@@ -429,9 +429,8 @@ static char *apply_prefix(char *buf, CHAR_DATA *ch, char *command) {
         return command;
     } else {
         int prefix_len = strlen(pc_prefix);
-        memcpy(buf, pc_prefix, (prefix_len < MAX_INPUT_LENGTH) ? prefix_len : MAX_INPUT_LENGTH - 1);
-        memcpy(buf + prefix_len, command, MAX_INPUT_LENGTH - prefix_len);
```
snellers, line 433: In debug builds, it'd bomb out on line 433 with a complaint about a stack overflow, and that cmd_buf underflows the "command" variable:

```
  This frame has 3 object(s):
    [32, 288) 'cmd_buf'
    [320, 576) 'command' <== Memory access at offset 288 partially underflows this variable
    [608, 864) 'logline'


```

snellers: I found that zeroing out the cmd_buf and using strncpy & strncat seemed to resolve it.  Thoughts?

snellers: Note: the 'at' command is a bit special as it parses its first arg (name of target character),  finds that, then sends the rest of your command string to interpret() to be interpreted like normal input.

snellers: Alright it seems that using strncpy() is itself nearly as bad, and won't compile on the newer compiler due to the size being copied is the same as the string length.

src/interp.c
```diff
@@ -445,7 +444,7 @@ void interpret(CHAR_DATA *ch, char *argument) {
     char command[MAX_INPUT_LENGTH];
     char logline[MAX_INPUT_LENGTH];
     const struct cmd_type *cmd;
-
+    memset(cmd_buf, 0, MAX_INPUT_LENGTH);
```
mattgodbolt, line 447: Zeroing it here means we must have been accessing it in an uninitialised state somehow before? Like, relying on whatever was there before, which is still kinda nasty. If this is part of the fix (and I don't dispute it stops the crash) it suggests something elsewhere in interpret() is reading past where it should (and now into the "safe space" of zeros you prepare)

src/interp.c
mattgodbolt: yikes yeah... this is super dodgy code (as it was before that is). I'm still thinking...

mattgodbolt:
Seems we converged on a solution, replacing the strlen/memcpy stuff with a single snprinf. I think it's best to revert the memset() too, it's FUDdish :)

PR #106, opened 2020-08-07
Descriptor refactor

mattgodbolt:
A partial refactor of the descriptors. It may be beneficial to go through the change commit-by-commit as I tried to be incremental.

Highlights:

* `Descriptor` class now holds the old descriptor info.
* `Descriptor`s are `new`/`deleted`. One day we'll do something different for the global descriptor list and be smart about memory etc, but we'll suffer this to get: constructor and destructor now do useful things.
* Descriptors no longer recycled in the free list. NB I am 90% sure it's leaking the "log_name" et al. A future change to `std::string` these will fix.
* Input handling is now more sane, uses dynamic lists of strings. I limit the size of incoming strings both by the doorman max line length, the MAX_INPUT in xania (as messages are dequeued), and by not allowing more than 50 (probably too high) queued messages.
* Input handling relies on doorman only ever sending whole lines. This seems safe and sensible.
* Most fields in Descriptor are still public. I plan on changing that up.

Thoughts welcomed. Obviously a ton of further improvements can be made, but this seemed like a good enough spot to stop and poll for comments and thoughts on direction.

This goes some way to mitigate #46 I think. Will do some more testing with that, but I couldn't get anything untoward happening with this change.

### DIFF
--- a/src/Descriptor.cpp
+++ b/src/Descriptor.cpp
@@ -1,4 +1,7 @@
-#include "DescriptorData.hpp"
+#include "Descriptor.hpp"
+
+#include "comm.hpp"
+#include "string_utils.hpp"
 
 // Up to 5 characters
 const char *short_name_of(DescriptorState state) {
@@ -65,10 +68,46 @@ Descriptor::~Descriptor() {
     /*    free_string(dclose->showstr_head); */
 }
 
-std::optional<std::string> Descriptor::pop_pending() {
+std::optional<std::string> Descriptor::pop_raw() {
     if (pending_commands_.empty())
         return std::nullopt;
     auto result = std::move(pending_commands_.front());
     pending_commands_.pop_front();
     return result;
+}
+
+std::optional<std::string> Descriptor::pop_incomm() {
+    auto maybe_next = pop_raw();
+    if (!maybe_next)
+        return {};
+
+    // Handle any backspace characters, newlines, or non-printing characters.
+    auto next_line = sanitise_input(*maybe_next);
+
+    if (next_line.size() > MAX_INPUT_LENGTH - 1) {
+        write("Line too long.\n\r");
+        return {};
+    }
+
+    // Do '!' substitution.
+    if (next_line[0] == '!')
+        return last_command_;
+    else {
+        last_command_ = next_line;
+        return next_line;
+    }
+}
+
+bool Descriptor::write(std::string_view text) const {
+    Packet p;
+    p.type = PACKET_MESSAGE;
+    p.channel = descriptor;
+
+    while (!text.empty()) {
+        p.nExtra = std::min<uint32_t>(text.length(), PACKET_MAX_PAYLOAD_SIZE);
+        if (!SendPacket(&p, text.data()))
+            return false;
+        text = text.substr(p.nExtra);
+    }
+    return true;
 }
--- a/src/Descriptor.hpp
+++ b/src/Descriptor.hpp
@@ -39,6 +39,9 @@ const char *name_of(DescriptorState state);
 struct Descriptor {
     static constexpr size_t MaxInbufBacklog = 50u;
     std::list<std::string> pending_commands_;
+    std::string last_command_;
+
+    [[nodiscard]] std::optional<std::string> pop_raw();
 
 public:
     Descriptor *next{};
@@ -52,8 +55,6 @@ public:
     DescriptorState connected{DescriptorState::GetName};
     uint16_t localport{};
     bool fcommand{};
-    char incomm[MAX_INPUT_LENGTH]{};
-    char inlast[MAX_INPUT_LENGTH]{};
     int repeat{};
     char *outbuf{};
     int outsize{2000};
@@ -63,9 +64,12 @@ public:
 
     explicit Descriptor(uint32_t descriptor);
     ~Descriptor();
+
     [[nodiscard]] bool is_playing() const noexcept { return connected == DescriptorState::Playing; }
     [[nodiscard]] bool is_input_full() const noexcept { return pending_commands_.size() >= MaxInbufBacklog; }
     void clear_input() { pending_commands_.clear(); }
     void add_command(std::string_view command) { pending_commands_.emplace_back(command); }
-    [[nodiscard]] std::optional<std::string> pop_pending();
+    [[nodiscard]] std::optional<std::string> pop_incomm();
+
+    bool write(std::string_view text) const;
 };
--- a/src/DescriptorData.hpp
+++ b/src/DescriptorData.hpp
@@ -2,6 +2,11 @@
 
 #include "merc.h"
 
+#include <list>
+#include <optional>
+#include <string>
+#include <string_view>
+
 // Connected state for a descriptor.
 enum class DescriptorState {
     Playing = 0,
@@ -32,6 +37,10 @@ const char *name_of(DescriptorState state);
  * Descriptor (channel) structure.
  */
 struct Descriptor {
+    static constexpr size_t MaxInbufBacklog = 50u;
+    std::list<std::string> pending_commands_;
+
+public:
     Descriptor *next{};
     Descriptor *snoop_by{};
     CHAR_DATA *character{};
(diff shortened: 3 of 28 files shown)


src/DescriptorData.hpp
```diff
@@ -3,26 +3,30 @@
...
-#define CON_GEN_GROUPS 11
-#define CON_PICK_WEAPON 12
-#define CON_READ_IMOTD 13
-#define CON_READ_MOTD 14
-#define CON_BREAK_CONNECT 15
-#define CON_GET_ANSI 16
-#define CON_CIRCUMVENT_PASSWORD 18 // used by doorman
-#define CON_DISCONNECTING 254 // disconnecting having been playing
-#define CON_DISCONNECTING_NP 255 // disconnecting before playing
+enum class DescriptorState {
```
snellers: aha, the C++ 11 enum class

src/Descriptor.hpp
```diff
@@ -0,0 +1,75 @@
...
+    char *host{};
+    char *logintime{};
+    uint32_t descriptor{};
+    uint32_t netaddr{};
+    DescriptorState connected{DescriptorState::GetName};
+    uint16_t localport{};
+    bool fcommand{};
+    int repeat{};
+    char *outbuf{};
+    int outsize{2000};
```
snellers, line 60: 2000 - arbitrary? It doesn't seem to have been a value we had before.

snellers: Ah nevermind I see it was in comm.c

src/Descriptor.cpp
```diff
@@ -0,0 +1,113 @@
...
+    outbuf = (char *)alloc_mem(outsize);
+    logintime = str_dup((char *)ctime(&current_time));
+    host = str_dup("(unknown)");
+}
+
+Descriptor::~Descriptor() {
+    free_string(host);
+    /* RT socket leak fix -- I hope */
+    free_mem(outbuf, outsize);
+    /*    free_string(dclose->showstr_head); */
```
snellers, line 68: As logintime is made with str_dup(), doesn't it also need to be free_string()d here?

src/db.cpp
```diff
@@ -2822,14 +2823,10 @@ void do_dump(CHAR_DATA *ch, char *argument) {
 
     /* descriptors */
     count = 0;
-    count2 = 0;
```
snellers, line 2825: This double counting was weird

src/comm.cpp
```diff
@@ -448,19 +448,20 @@ void game_loop_unix(int control) {
             /* Waitstate the character */
             if (d->character && d->character->wait)
                 continue;
-            read_from_buffer(d);
-            if (d->incomm[0] != '\0') {
+
+            if (auto incomm = d->pop_incomm(); incomm) {
```
snellers, line 452: std::optional's syntax in an if is kind of weird but makes sense :)

snellers: Nice to see that you were able to ditch the copying here. It seemed...unnatural.

src/comm.cpp
```diff
@@ -639,121 +618,37 @@ void close_socket(DESCRIPTOR_DATA *dclose) {
...
-   /* Hold horses if pending command already. */
-   if ( d->incomm[0] != '\0' )
-      return true;
-#endif
-
-    /* Check for overflow. */
-    iStart = strlen(d->inbuf);
-    if (iStart >= (int)sizeof(d->inbuf) - 10) {
+bool read_from_descriptor(Descriptor *d, char *data, int nRead) {
+    if (d->is_input_full()) {
```
snellers, line 626: Much nicer!

mattgodbolt: I just hope it's right :)

src/comm.cpp
mattgodbolt: Yay! Thanks :) In fact the `; incomm)` is unnecessary; the `if()` doesn't need it in this case (it was refactored from when it used to be `...; !incomm.empty())`

src/Descriptor.cpp
mattgodbolt: Absolutely! That's one of the leaks we're suffering from I mentioned in the notes. Rather than fix it here, I'll make all of them `std::string` in a followup. I was somewhat arbitrary in the line-drawing between something worthwhile to review vs a gargantuan thread-pulling mess :)

src/DescriptorData.hpp
mattgodbolt: Yes! Not quite as good as Java's equivalent, unfortunately.

src/Descriptor.cpp
mattgodbolt: It's another reminder that all the object and string pooling is hiding any real memory leaks from leak sanitizer/valgrind et al...we'll get there :)

src/string_utils.cpp
```diff
@@ -79,3 +79,24 @@ std::string remove_last_line(std::string_view str) {
...
+        if (c == '\b') {
+            if (!result.empty())
+                result.pop_back();
+        } else {
+            result.push_back(c);
+        }
+    }
+
+    // Remove non-printing, non-ascii characters. This awkward erase(remove_if,,,) idiom is a uniquely C++ mess.
+    result.erase(std::remove_if(result.begin(), result.end(), [](char c) { return !isascii(c) || !isprint(c); }),
```
snellers, line 98: ah yes I remember running into this a while back too...clunky but ok once you know it :)

mattgodbolt: Yeah, coming "sometime soon" is C++ ranges which will hopefully make this better...

src/Descriptor.cpp
```diff
@@ -0,0 +1,113 @@
...
+    case DescriptorState::CircumventPassword: return "CPass";
+    case DescriptorState::Disconnecting: return "Disc";
+    case DescriptorState::DisconnectingNp: return "DiscN";
+    }
+    return "<UNK>";
+}
+
+const char *name_of(DescriptorState state) {
+    switch (state) {
+    case DescriptorState::Playing: return "Playing";
```
philjessies, line 34: Personally, I feel this is one of those rare moments when a macros makes the code better:

`#define CS(x) case DescriptorState::x: return #x;`

That way, you can't accidentally end up with something inconsistent (by copy/paste/imperfect-hacking).

That said, this entire function isn't used anywhere, so I'd suggest killing it. If the short name is all we need, then let's go with just that.

src/Descriptor.cpp
```diff
@@ -0,0 +1,113 @@
...
+    case DescriptorState::GetAnsi: return "GetAnsi";
+    case DescriptorState::CircumventPassword: return "CircumventPassword";
+    case DescriptorState::Disconnecting: return "Disconnecting";
+    case DescriptorState::DisconnectingNp: return "DisconnectingNp";
+    }
+    return "Unknown";
+}
+
+Descriptor::Descriptor(uint32_t descriptor) {
+    this->descriptor = descriptor;
```
philjessies, line 58: Shouldn't this be done in the initialiser list?

src/Descriptor.hpp
```diff
@@ -0,0 +1,75 @@
...
+struct Descriptor {
+    static constexpr size_t MaxInbufBacklog = 50u;
+    std::list<std::string> pending_commands_;
+    std::string last_command_;
+
+    [[nodiscard]] std::optional<std::string> pop_raw();
+
+public:
+    Descriptor *next{};
+    Descriptor *snoop_by{};
```
philjessies, line 48: Dunno if we're doing TODOs, but if so, maybe having a couple of them here would be good. IMO it'd be good to:
* Switch these fields to 'private'.
* Add a trailing _ (IIRC that was the style we ended up agreeing on, right?)

src/Descriptor.cpp
mattgodbolt: Yeah, agreed. I sometimes use macros; but IDE support makes this pretty easy (refactors rename the strings too). Agreed re:killing it off, it's habit and a hedge for a future world in which I want to make sure (for debugging) we can stream out any enums etc. But...will remove.

src/Descriptor.cpp
mattgodbolt: 100%! This is what happens when you paste code from a `new_function` and "just" replace `pnew` with `this` :D

mattgodbolt: THOUGH! one tricky point: outsize is initialized _after_ outbuf. stupid C++, so I'll have to switch the order of those in the struct too.

mattgodbolt: (Also these are all going soon...)

src/Descriptor.hpp
mattgodbolt: Yes! all these are planned! Just one small step at a time (see the upcoming PR that does some of these).  :D

PR #107, opened 2020-08-07
Thread-pulling in descriptor:

mattgodbolt:
* make hostname and login time std:strings
* following on the chain of nastiness, revamp the info system a tiny bit.
* Fixes a niggling issue where `finger themoog` shows "Themoog" instead of "TheMoog" for case-magic users.
* sketch a FD wrapper, and use it in info.cpp to fix a FD leak
* unify implementation (with a TODO) of redaction, and redact on changes only.


mattgodbolt:
Sorry this is so large, in particular the changes in info.cpp are gnarly. But it's a net code removal.

src/CMakeLists.txt
```diff
@@ -56,8 +56,9 @@ add_library(xania_lib STATIC
         wiznet.cpp
         xania.cpp
         Descriptor.cpp Descriptor.hpp
-        string_utils.cpp string_utils.hpp)
-target_link_libraries(xania_lib crypt chat)
+        string_utils.cpp string_utils.hpp
+        info.hpp WrappedFd.cpp WrappedFd.hpp)
+target_link_libraries(xania_lib crypt chat CONAN_PKG::fmt)
```
snellers, line 61: ok so an actual fmt shared lib we're linking with

src/Descriptor.cpp
```diff
@@ -111,3 +112,26 @@ bool Descriptor::write(std::string_view text) const {
...
+// users coming from the same IP.
+std::string get_masked_hostname(std::string_view hostname) {
+    return "{}*** [#{}]"_format(hostname.substr(0, 6), djb2_hash(hostname));
+}
+
+}
+
+void Descriptor::raw_full_hostname(std::string_view raw_full_hostname) {
+    raw_host_ = raw_full_hostname;
+    redacted_host_ = get_masked_hostname(raw_host_);
```
snellers, line 136: masked/redacted,  should probably use the same word everywhere

src/WrappedFd.hpp
```diff
@@ -0,0 +1,31 @@
...
+            fclose(fd_);
+        fd_ = nullptr;
+    }
+
+public:
+    explicit WrappedFd(FILE *fd) : fd_(fd) {}
+    WrappedFd(const WrappedFd &) = delete;
+    WrappedFd &operator=(const WrappedFd &) = delete;
+    WrappedFd(WrappedFd &&other) noexcept : fd_(std::exchange(other.fd_, nullptr)) {}
+    WrappedFd &operator=(WrappedFd &&other) noexcept {
```
snellers, line 21: It makes me wonder if we'd ever actually do assignment of these :)

src/WrappedFd.hpp
```diff
@@ -0,0 +1,31 @@
...
+        close();
+        fd_ = std::exchange(other.fd_, nullptr);
+        return *this;
+    }
+    ~WrappedFd() { close(); }
+    operator bool() const noexcept { return fd_ != nullptr; }
+    operator FILE *() const noexcept { return fd_; }
+
+    static WrappedFd open_text(const std::string &filename) { return WrappedFd(fopen(filename.c_str(), "r")); }
```
snellers, line 30: not sure I get the _text nomenclature

src/CMakeLists.txt
mattgodbolt: Right! 99.9% of fmt is headers but there's a nugget that's a library. Either way, to get access to the headers, we "link" against it. CMake then arranges for the right `-Ipath/to/fmt/headers`

src/info.cpp
```diff
@@ -35,16 +38,20 @@ void read_char_info(FINGER_INFO *info);
         break;                                                                                                         \
     }
 
-/* For finger info */
-FINGER_INFO *info_cache = nullptr;
+// For finger info.
+namespace {
```
snellers, line 42: I notice you've namespaced these new sections of act_info.cpp like here and down on line 76, wondering if they can be consolidated together

src/WrappedFd.hpp
mattgodbolt: It's easy to do by mistake, if you accidentally wrote something that took a WrappedFd by value:

```void print_debug_to(WrappedFd fd);```

you'd double-close the FD... (the parameter itself is a copy of the passed-in value...ugh)

Also in general there's the "rule of 5" - if you define ANY of the special functions, define all of them. The table of which defaults get applied is COMPLICATED so best to spell out what you want.

File: src/Descriptor.cpp (hunk 2 above)
Comment (mattgodbolt) [reply src/Descriptor.cpp:136]: 100%. changing to masked throughout

File: src/WrappedFd.hpp (hunk 4 above)
Comment (mattgodbolt) [reply src/WrappedFd.hpp:30]: oh! as `text` ("r") vs `binary` ("rb"). We probably won't ever open binary so I'll rename to `open`

File: src/info.cpp (hunk 5 above)
Comment (mattgodbolt) [reply src/info.cpp:42]: Can do: I was (for now) sticking to keeping all the heinous globals bunched up together at the top. I think it's cool to move them in this case, there aren't too many. Also I don't actually know if `player_list` is used meaningfully anywhere...I'll add a TODO and issue to track that

File: src/merc.h [hunk 6]
```diff
@@ -85,7 +86,6 @@ typedef struct shop_data SHOP_DATA;
 typedef struct time_info_data TIME_INFO_DATA;
 typedef struct weather_data WEATHER_DATA;
 /* Rohan's finger caching struct */
```
snellers, line 88: --comment

mattgodbolt: I felt bad taking them out .. but yeah. The time has come :)

mattgodbolt: oh, you meant for finger_caching...yeah done!

PR #112, opened 2020-08-08
Descriptor refactor: snooping

mattgodbolt:
Snooping now works on sets. Multisnooping now allowed. Hopefully clearer code throughout.

### DIFF
--- a/src/Descriptor.cpp
+++ b/src/Descriptor.cpp
@@ -34,6 +34,11 @@ const char *short_name_of(DescriptorState state) {
 
 Descriptor::Descriptor(uint32_t descriptor) : channel_(descriptor), login_time_(ctime(&current_time)) {}
 
+Descriptor::~Descriptor() {
+    // Ensure we don't have anything pointing back at us. No messages here in case this is during shutdown.
+    stop_snooping();
+}
+
 std::optional<std::string> Descriptor::pop_raw() {
     if (pending_commands_.empty())
         return std::nullopt;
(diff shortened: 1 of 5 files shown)


src/Descriptor.cpp
```diff
@@ -105,10 +110,12 @@ bool Descriptor::flush_output() noexcept {
     if (outbuf_.empty())
         return true;
 
-    if (snoop_by && character) {
-        snoop_by->write(character->name);
-        snoop_by->write("> ");
-        snoop_by->write(outbuf_);
+    if (character) {
+        for (auto *snooper : snoop_by_) {
+            snooper->write(character->name);
```
mattgodbolt, line 115: Occurs to me I can use the new for line iterator to prefix each _line_ with `char> ` instead of just the first bit. Might make it clearer. Not that we use snoop a whole lot!

snellers: Currently it's a 95+ command. I don't think it should be used casually except in testing. If the mud ends up being played by more people then it should only be used if it looks like someone's obviously breaking rules and you can't tell by standing there while wizinvis. And perhaps then only after they've already been warned. Or if another player has complained about their behavior.

src/Descriptor.cpp
```diff
@@ -151,3 +158,53 @@ void Descriptor::show_next_page(std::string_view input) noexcept {
...
+
+void Descriptor::close() noexcept {
+    while (!snoop_by_.empty()) {
+        auto &snooper = *snoop_by_.begin();
+        snooper->write("Your victim ({}) has left the game.\n\r"_format(character ? character->name : "unknown"));
+        snooper->stop_snooping(*this);
+    }
+    stop_snooping();
+}
+
```
snellers, line 203: I think there may alternatives for 'note_input' that could be clearer, e.g. notify_snoopers? Your call.

mattgodbolt: Yeah, I went back and forth here. `notify_snoopers()` is defininetly what it does _now_, but at some point in the future I can see that other stuff could be done here (logging maybe?). So I hedged my bets as if it were a general purpose "listener". For now I think I'll keep that; and then if we come across it again and it's still only doing snooping we can rename. thanks!

PR #113, opened 2020-08-08
Add a new CommandSet class, to take over the work of the 'trie', but …

philjessies:
…in a smaller, better, and more unit-tested way. So far I've only converted the 'note' subcommands - the others will follow, assuming this is OK.

### DIFF
--- a/src/CommandSet.hpp
+++ b/src/CommandSet.hpp
@@ -1,50 +1,54 @@
 #pragma once
 
+#include "string_utils.hpp"
 #include <functional>
 #include <map>
 #include <string>
 #include <utility>
-#include "string_utils.hpp"
 
-template<typename T>
+template <typename T>
 class CommandSet {
- public:
-  CommandSet() {}
-  
-  // add a command called 'name' with a value, which is available to characters
-  // at level 'level' or above.
-  void add(const std::string& name_in, T value, int level) {
-    auto name = lower_case(name_in);
-    commands_[name] = std::make_pair(value, level);
-  }
-  
-  std::optional<const T> get(const std::string& name_in, int level) const {
-    auto name = lower_case(name_in);
-    // lower_bound gets the lowest item whose key is great than or equal to name.
-    for (auto it = commands_.lower_bound(name); it != commands_.end(); it++) {
-      // If this command is too high a level for the user, skip on to the next one,
-      // so that whatever level the user is, there are no weird gaps that might give
-      // a hint as to the imm commands.
-      if (it->second.second > level) continue;
-      // This command is the right level for the user. However, we may have skipped past
-      // the first valid prefix, or indeed there may be no command with a valid prefix.
-      // So if this item doesn't match the requested key, drop out to the 'null' return.
-      if (!has_prefix(it->first, name)) break;
-      // If we got here, the user is of the right level to use the command, and the command
-      // matches what was requested.
-      return it->second.first;
+public:
+    CommandSet() {}
+
+    // add a command called 'name' with a value, which is available to characters
+    // at level 'level' or above.
+    void add(const std::string &name_in, T value, int level) {
+        auto name = lower_case(name_in);
+        commands_[name] = std::make_pair(value, level);
     }
-    return std::nullopt;
-  }
-  
-  void enumerate(int min_level, int max_level, std::function<void(const std::string& name, const T value, int level)> fn) const {
-    for (auto it = commands_.begin(); it != commands_.end(); it++) {
-      auto level = it->second.second;
-      if (level < min_level || level > max_level) continue;
-      fn(it->first, it->second.first, level);
+
+    std::optional<const T> get(const std::string &name_in, int level) const {
+        auto name = lower_case(name_in);
+        // lower_bound gets the lowest item whose key is great than or equal to name.
+        for (auto it = commands_.lower_bound(name); it != commands_.end(); it++) {
+            // If this command is too high a level for the user, skip on to the next one,
+            // so that whatever level the user is, there are no weird gaps that might give
+            // a hint as to the imm commands.
+            if (it->second.second > level)
+                continue;
+            // This command is the right level for the user. However, we may have skipped past
+            // the first valid prefix, or indeed there may be no command with a valid prefix.
+            // So if this item doesn't match the requested key, drop out to the 'null' return.
+            if (!has_prefix(it->first, name))
+                break;
+            // If we got here, the user is of the right level to use the command, and the command
+            // matches what was requested.
+            return it->second.first;
+        }
+        return std::nullopt;
     }
-  }
-  
- private:
-  std::map<std::string, std::pair<T, int>> commands_;
+
+    void enumerate(int min_level, int max_level,
+                   std::function<void(const std::string &name, const T value, int level)> fn) const {
+        for (auto it = commands_.begin(); it != commands_.end(); it++) {
+            auto level = it->second.second;
+            if (level < min_level || level > max_level)
+                continue;
+            fn(it->first, it->second.first, level);
+        }
+    }
+
+private:
+    std::map<std::string, std::pair<T, int>> commands_;
 };
--- a/src/CommandSet.hpp
+++ b/src/CommandSet.hpp
@@ -9,16 +9,14 @@
 template <typename T>
 class CommandSet {
 public:
-    CommandSet() {}
-
     // add a command called 'name' with a value, which is available to characters
     // at level 'level' or above.
-    void add(const std::string &name_in, T value, int level) {
+    void add(const std::string &name_in, T &&value, int level) {
         auto name = lower_case(name_in);
-        commands_[name] = std::make_pair(value, level);
+        commands_.try_emplace(name, std::move(value), level);
     }
 
-    std::optional<const T> get(const std::string &name_in, int level) const {
+    std::optional<T> get(const std::string &name_in, int level) const {
         auto name = lower_case(name_in);
         // lower_bound gets the lowest item whose key is great than or equal to name.
         for (auto it = commands_.lower_bound(name); it != commands_.end(); it++) {
@@ -39,14 +37,19 @@ public:
         return std::nullopt;
     }
 
-    void enumerate(int min_level, int max_level,
-                   std::function<void(const std::string &name, const T value, int level)> fn) const {
+    void enumerate(std::function<void(const std::string &name, T value, int level)> fn) const {
         for (auto it = commands_.begin(); it != commands_.end(); it++) {
-            auto level = it->second.second;
-            if (level < min_level || level > max_level)
-                continue;
-            fn(it->first, it->second.first, level);
+            fn(it->first, it->second.first, it->second.second);
         }
+    }
+
+    std::function<void(const std::string &name, T value, int level)>
+    level_restrict(int min_level, int max_level,
+                   std::function<void(const std::string &name, T value, int level)> fn) const {
+        return [min_level, max_level, fn](const std::string &name, T value, int level) {
+            if (level >= min_level && level <= max_level)
+                fn(name, value, level);
+        };
     }
 
 private:
--- a/src/CommandSet.hpp
+++ b/src/CommandSet.hpp
@@ -1,0 +1,50 @@
+#pragma once
+
+#include <functional>
+#include <map>
+#include <string>
+#include <utility>
+#include "string_utils.hpp"
+
+template<typename T>
+class CommandSet {
+ public:
+  CommandSet() {}
+  
+  // add a command called 'name' with a value, which is available to characters
+  // at level 'level' or above.
+  void add(const std::string& name_in, T value, int level) {
+    auto name = lower_case(name_in);
+    commands_[name] = std::make_pair(value, level);
+  }
+  
+  std::optional<const T> get(const std::string& name_in, int level) const {
+    auto name = lower_case(name_in);
+    // lower_bound gets the lowest item whose key is great than or equal to name.
+    for (auto it = commands_.lower_bound(name); it != commands_.end(); it++) {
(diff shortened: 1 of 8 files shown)


src/CommandSet.hpp
```diff
@@ -0,0 +1,50 @@
...
+      // So if this item doesn't match the requested key, drop out to the 'null' return.
+      if (!has_prefix(it->first, name)) break;
+      // If we got here, the user is of the right level to use the command, and the command
+      // matches what was requested.
+      return it->second.first;
+    }
+    return std::nullopt;
+  }
+  
+  void enumerate(int min_level, int max_level, std::function<void(const std::string& name, const T value, int level)> fn) const {
```
snellers, line 40: there's probably a fancy way to do this in a functional style with a lambda & a filter on the levels but I'm too rusty to remember how without experimenting :)

src/note.cpp
```diff
@@ -8,24 +8,25 @@
...
+#include <functional>
+#include <memory>
 
 static NOTE_DATA *note_first;
 static NOTE_DATA *note_last;
 
-static void *trie;
-
-typedef void(note_fn_t)(CHAR_DATA *, const char *);
+typedef std::function<void(CHAR_DATA *ch, const char *arg)> note_fn;
```
snellers, line 28: note_fn could potentially reuse the one that's currently in interp.cpp line 45 as it's the same signature.

src/CommandSet.hpp
```diff
@@ -0,0 +1,50 @@
...
+  // at level 'level' or above.
+  void add(const std::string& name_in, T value, int level) {
+    auto name = lower_case(name_in);
+    commands_[name] = std::make_pair(value, level);
+  }
+  
+  std::optional<const T> get(const std::string& name_in, int level) const {
+    auto name = lower_case(name_in);
+    // lower_bound gets the lowest item whose key is great than or equal to name.
+    for (auto it = commands_.lower_bound(name); it != commands_.end(); it++) {
```
mattgodbolt: There's a cute thing for this: [equal_range](https://en.cppreference.com/w/cpp/container/map/equal_range) returns a pair of iterators that bound the range. _Maybe_ useful here?

```
auto &&[first, last] = commands_.equal_range(name);
for (auto it = first; it != last; ++it)
```

I'm not 100% sure that's the same as what you're doing though!

src/CommandSet.hpp
```diff
@@ -0,0 +1,50 @@
...
+  CommandSet() {}
+  
+  // add a command called 'name' with a value, which is available to characters
+  // at level 'level' or above.
+  void add(const std::string& name_in, T value, int level) {
+    auto name = lower_case(name_in);
+    commands_[name] = std::make_pair(value, level);
+  }
+  
+  std::optional<const T> get(const std::string& name_in, int level) const {
```
mattgodbolt: So: here you're returning a copy of the T you were given. Making it `const` is just annoying for the caller: it's _their_ T now! so: I'd suggest `optional<T>` here.

SUPER SUPER ANNOYINGLY you can't have what you probably want, which is `optional<const T&>`. Debate rages in the C++ community over this. 

If you want to avoid a copy, then returning a `const T*` and `nullptr` instead is your best choice :(

src/CommandSet.hpp
mattgodbolt: Nice! std::function makes it super clear what the signature of the function is :)

I'm lazy and typically do:
```
template<typename Func>
void enumerate(int min_level, int max_level, const Func &fn) { ...same...}
```

It's lazy as "anything" can go in the Func. It has the advantage of allowing naked functions, or function objects or anything Callable here instead of forcing the caller to make (or the compiler to do the work to construct) a std::function. Just heads up!

src/note.cpp
mattgodbolt, line 28: `typedef` is the Old Way™. It still works (as everything does in C++), but it's nicer to use:

```cpp
using note_fn = std::function<void(CHAR_DATA *ch, const char *arg)>;
```

(mainly as that syntax works in more cases)

src/note.cpp
```diff
@@ -8,24 +8,25 @@
...
+static std::unique_ptr<CommandSet<note_fn>> sub_commands;
```
mattgodbolt, line 29: Here why not "just" have an object. It's a global, and yes, construction order is painful, but that's only painful when an object refers to other objects. This does not; so you can just make this `CommandSet<note_fn> sub_commands`

src/note.cpp
```diff
@@ -503,22 +502,17 @@ static void note_readfile() {
...
-    trie_add(trie, "post", (void *)&note_post, 0);
-    trie_add(trie, "send", (void *)&note_post, 0);
-    trie_add(trie, "remove", (void *)&note_removecmd, 0);
-    trie_add(trie, "delete", (void *)&note_delete, MAX_LEVEL - 2);
+    sub_commands = std::make_unique<CommandSet<note_fn>>();
+    sub_commands->add("read", note_read, 0);
+    sub_commands->add("list", note_list, 0);
+    sub_commands->add("+", note_addline, 0);
+    sub_commands->add("-", note_removeline, 0);
+    sub_commands->add("subject", note_subject, 0);
```
mattgodbolt, line 510: Yay death to TODOs

src/string_utils.cpp
```diff
@@ -120,3 +120,22 @@ impl::LineSplitter::Iter impl::LineSplitter::Iter::operator++() noexcept {
     *this = Iter(rest_, false);
     return prev;
 }
+
+std::string lower_case(std::string_view str) {
```
mattgodbolt, line 124: Hurrah!

src/string_utils.cpp
```diff
@@ -120,3 +120,22 @@ impl::LineSplitter::Iter impl::LineSplitter::Iter::operator++() noexcept {
...
+    std::string result;
+    result.reserve(str.size());
+    for (auto c : str) {
+        result.push_back(tolower(c));
+    }
+    return result;
+}
+
+bool has_prefix(std::string_view haystack, std::string_view needle) {
+    if (needle.size() > haystack.size())
```
mattgodbolt, line 134: I _think_ this is the same as:

```
return haystack.find(needle) == 0;
```

right? or:

```
return needle.size() <= haystack.size() || haystack.substr(0, needle.size() != needle;
```

Either way, the loop here I think can be replaced with a string funciont. C++ doesn't have many (as you've noticed)... but there _are_ some...

src/test/CommandSetTest.cpp
```diff
@@ -0,0 +1,86 @@
+#include "CommandSet.hpp"
```
mattgodbolt, line 1: I am SO SO SO happy to see this! THANK YOU :)

src/test/CommandSetTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+TEST_CASE("CommandSet tests") {
+    SECTION("empty") {
+        CommandSet<int> cs;
+        CHECK(cs.get("nothing", 1).has_value() == false);
+        CHECK(enumerate_str(cs, 0, 100) == "");
+    }
+    SECTION("level limits") {
+        CommandSet<std::string> cs;
+        cs.add("kick", "foot", 0);
```
mattgodbolt, line 25: hahah

src/test/CommandSetTest.cpp
```diff
@@ -0,0 +1,86 @@
+#include "CommandSet.hpp"
+
+#include <catch2/catch.hpp>
+
+#include <sstream>
+#include <string_view>
+
+template <typename T>
+std::string enumerate_str(const CommandSet<T> &cs, int min_level, int max_level) {
```
mattgodbolt, line 9: Love it!
Would you mind wrapping it in an anonymous namespace (the "improved" `static`)... just reduces the chance of two files defining the same function by accident.

src/test/string_utils_test.cpp
```diff
@@ -137,4 +137,19 @@ TEST_CASE("string_util tests") {
             CHECK(split_lines<vs>("first\r\n\r\nthird\r\n") == vs{"first", "", "third"});
         }
     }
+
+    SECTION("lower_case") {
+        SECTION("already lower") { CHECK(lower_case("already lowercase") == "already lowercase"); }
+        SECTION("all upper") { CHECK(lower_case("I AM SHOUTY") == "i am shouty"); }
+        SECTION("mixture") { CHECK(lower_case("I am A LITTLE shouty") == "i am a little shouty"); }
+        SECTION("non-letters") { CHECK(lower_case("G0RD0N'5 AL111111V3!!!!!!") == "g0rd0n'5 al111111v3!!!!!!"); }
```
mattgodbolt, line 145: lol

src/test/string_utils_test.cpp
```diff
@@ -137,4 +137,19 @@ TEST_CASE("string_util tests") {
...
+    }
+
+    SECTION("has_prefix") {
+        SECTION("prefix") { CHECK(has_prefix("boolean", "bool") == true); }
```
mattgodbolt, line 149: I much prefer not `== false` or `==true` here: I find it visually noisy...

Also CHECK() is a magic macro that tries to understand and print out the intention of a (failing) test. It's smart enough to know if something was a bool check and prints something nice in that case. Instead of "expected 'true' got 'false'", which is...sort of redundant.

Minor gripe anyway! I know other languages are big on `== true` or `!= nullptr` but C++ idiomatically (IME!) isn't like that.

src/note.cpp
snellers: right, and we already have that defined in interp

philjessies: Done.

philjessies: Similar signature, yes (const here, non-const in interp.cpp). To reuse it, though, I'd have to move it to a .h file. Plus, I think that while they do have very similar signatures now, they don't necessarily refer to the same thing.

src/note.cpp
philjessies: Done.

src/note.cpp
philjessies: TODOs can also shake hands with Death.

src/string_utils.cpp
philjessies: Yeah, why on earth does STL not already have this?

src/string_utils.cpp
philjessies: Yeah, actually I'd rather do the latter. 'find' is way more expensive*. Done.

* yeah. Noticably so. Sure.

src/test/CommandSetTest.cpp
philjessies: no trouble. I've not touched C++ properly for ages; it's great practice.

src/test/CommandSetTest.cpp
philjessies: good point. Plus, it doesn't destroy all debug stuff.

src/test/string_utils_test.cpp
philjessies: Fair enough. I don't feel strongly either way, so switched to ditch the == true and == false.

src/test/CommandSetTest.cpp
mattgodbolt: > good point. Plus, it doesn't destroy all debug stuff.

Which what? :)

philjessies: Oh, was thinking of in real code: 'static' symbols tend to get ditched by the compiler, but this is a unit test, so not really important here.

mattgodbolt: Right. `static` means the link-level symbols are not defined, yes. But (in theory) not lost from the debug symbols. I tend to build everything with debug symbols on (even release builds) for this reason!

src/string_utils.cpp
mattgodbolt: Neat! find does look for it _anywhere_ and I suspect the compiler isn't smart enough to work out it can stop early. (if only there was some interactive website where we could find out...)

src/CommandSet.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <functional>
+#include <map>
+#include <string>
+#include <utility>
+#include "string_utils.hpp"
+
+template<typename T>
+class CommandSet {
+ public:
+  CommandSet() {}
```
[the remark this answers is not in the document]
philjessies: Done.

src/CommandSet.hpp
```diff
@@ -0,0 +1,50 @@
...
+  
+  // add a command called 'name' with a value, which is available to characters
+  // at level 'level' or above.
+  void add(const std::string& name_in, T value, int level) {
+    auto name = lower_case(name_in);
+    commands_[name] = std::make_pair(value, level);
```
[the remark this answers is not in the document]
philjessies: failing on inserting dups is good here. Changed as suggested - your syntax was bang on.

src/CommandSet.hpp
philjessies: Oh, that works. I was expecting it to refuse to let me return a non-const T, because this is a const function. But I suppose as it's theoretically returning-by-value (despite the fact that T may be a pointer, reference or whatever) it makes sense to be able to return std::optional<T>.

src/CommandSet.hpp
philjessies: From what I can see, that's just a way of getting lower_bound and upper_bound together. Whereas I want entries that the query is a prefix of (which would be outside the 'equal range'). With the proviso that I only match stuff which is available to the player's level. So it doesn't look like equal_range is really what I want. I guess it's intended for std::multimap etc.

src/CommandSet.hpp
philjessies: OK, I've got it figured out, I think. Created a level_restrict function which filters the function with the level checks. Should be OK; I think the filtering's only needed in a few places anyway.

PR #116, opened 2020-08-09
Cleanup chat

snellers:


snellers:
As I was saying: this is mainly about renaming stuff. The old code was super hard to follow. I've gone through tons of the variables and methods to give them names that mean something, as well as strip out some cruft. This opens the door to some more modernizations. But parts of the code will be hard to rewrite to use std:; as it's so gnarly.

src/act_comm.cpp
```diff
@@ -941,24 +941,28 @@ bool is_same_group(CHAR_DATA *ach, CHAR_DATA *bch) {
...
+ * Trigger the Eliza chat engine. Can cause an NPC to say something back to the player.
+ * to_npc: the NPC that received the chat/social message.
+ * from_player: the player that sent it.
+ */
+void chatperform(CHAR_DATA *to_npc, CHAR_DATA *from_player, const char *msg) {
+    char response_buf[MaxChatReplyLength];
+    if (!IS_NPC(to_npc) || (from_player != nullptr && IS_NPC(from_player)))
         return; /* failsafe */
-    reply = dochat(ch->name, msg, victim ? victim->name : "you");
+    const char *reply = dochat(response_buf, from_player ? from_player->name : "you", msg, to_npc->name);
```
snellers, line 953: It was using a static char array to store the response, I ditched that. And I flipped the player/npc args around.

src/chat/Database.cpp
```diff
@@ -0,0 +1,38 @@
+// allkeys.cpp
+// chris busch
+
+#include "Database.hpp"
+
+namespace chat {
```
snellers, line 6: Database is the new "allkeys"

src/chat/Database.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace chat {
+
+/**
+ * Databases may be configured to link to one other database.
+ * This indicates that a Database does not have a link.
+ */
+constexpr inline auto NoDatabaseLink = -1;
+
+/**
```
snellers, line 13: I'm hoping to kill Record and replace it with a vector of the KeywordResponses

src/chat/Eliza.cpp
```diff
@@ -0,0 +1,421 @@
...
+            ++i;
+            if (point == toofar) {
+                *point = '\0';
+                return;
+            }
+        }
+    }
+    *point++ = '\0';
+}
+
```
snellers, line 208: This guy and match() are the real meat of the code. Lots of renaming here.

src/chat/KeywordResponses.cpp
```diff
@@ -0,0 +1,48 @@
+// akey.cpp
+
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+
```
snellers, line 6: KeywordResponses is the new "akey"

src/chat/KeywordResponses.cpp
```diff
@@ -0,0 +1,48 @@
...
+    for (int i = 0; i < num_responses_; i++) {
+        if ((total += (responses_[i].weight_)) > rnd_num) {
+            return responses_[i].message_;
+        }
+    }
+    return defaultmsg;
+}
+
+int KeywordResponses::add_keywords(char *keyword) { return (keywords_ = strdup(keyword)) != nullptr; }
+
```
snellers, line 31: I'd like to get rid of this nasty alloc business.  Thinking vector.

src/chat/KeywordResponses.hpp
```diff
@@ -0,0 +1,38 @@
+// chris busch
+
+#pragma once
+
+namespace chat {
+
```
snellers, line 6: There were confusing misspellings in the old code like "wate". After poking around I realised it uses weights when calculating which response to fire back.

src/chat/Database.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+Record *Database::add_record() {
+    Record *record;
+    if ((record = new Record) == nullptr) {
```
mattgodbolt, line 10: per other comments, we try our best to not use raw pointers in modern C++. Again, probably a step on the way toward that :)

`new` cannot return `nullptr`. Might as well get rid of that, if nothing else :)

src/chat/Database.cpp
```diff
@@ -0,0 +1,38 @@
...
+        top_->next_ = record;
+    current_ = top_ = record;
+    top_->next_ = nullptr;
+    if (record_count_ == 0)
+        first_ = top_;
+    record_count_++;
+    return current_;
+}
+
+Record *Database::reset() { return (current_ = first_); }
```
mattgodbolt, line 23: be aware the parens here are extraneous _and actually change the meaning in some (odd) contexts_. In general I'd avoid them. Specifically in `return` statements.

Also the assignment in a return itself is pretty uncommon and surprising. A linter may explode here as it looks like a comparison.

If nobody uses the return value, I'd drop it.

src/chat/Database.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+Record *Database::next_record() {
+    if (current_ != nullptr) {
```
mattgodbolt, line 26: Stylistically it's very common in C and C++ to not compare with `nullptr` or similar. Sorry if this is naggy, and/or probably just copied from the eliza OG code.

src/chat/Database.hpp
mattgodbolt: Cool! Thanks. If you keep the poitner stuff (you probably won't) be aware of pointer stability.

src/chat/Database.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * the name of an NPC a player talks to or interacts with.
+ *
+ * A database can optionally be linked to one other database using the @ syntax in the file.
+ * This provides a way to share database records between databases.
+ */
+class Database {
+public:
+    int linked_database_num;
+    Database() : linked_database_num(NoDatabaseLink) {}
+    Record *current_record() { return current_; }
```
mattgodbolt, line 34: you've probably seen the chatty `[[nodiscard]]` crap in my code: it's to note those things for which you probably should use the return value. Feasibly some of these need it? e.g. `add_record()` ?

src/chat/Eliza.cpp
```diff
@@ -0,0 +1,421 @@
...
+#include "chatconstants.hpp"
+
+using namespace chat;
+
+/////YOU MAY NOT change the next 2 lines.
+const char eliza_title[] = "chat by Christopher Busch  Copyright (c)1993";
+const char eliza_version[] = "version 1.0.0";
+
+// Gets a word out of a string.
+int Eliza::get_word(const char *&input, char *outword, char &outother) {
```
mattgodbolt, line 23: (fyi the `const char *&` is a surprising thing to callers not expecting their params to get fiddled with. Probably from the OG.

src/chat/Eliza.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+void Eliza::sort_databases_by_name() {
+    qsort((char *)named_databases_, num_names_, sizeof(Eliza::NamedDatabase), compare_database_name);
+}
+
+int Eliza::get_database_num_by_exact_name(const char *n) {
+    NamedDatabase finder;
+    if (finder.set(n, 0) == nullptr)
+        return 0;
+    NamedDatabase *found = (NamedDatabase *)bsearch((char *)&finder, (char *)named_databases_, num_names_,
```
mattgodbolt, line 66: crying out for some STL :)

src/chat/Eliza.cpp
mattgodbolt: Looks so much cleaner already!

src/chat/Eliza.cpp
```diff
@@ -0,0 +1,421 @@
...
+    }
+    fclose(data);
+    if (!recurflag) {
+        num_databases_++;
+        sort_databases_by_name();
+    }
+    return true;
+}
+
+char *Eliza::trim(char str[]) {
```
mattgodbolt, line 313: Would be ace to rephrase these using the string_utils. I think this does a two-sided trim? We don't have that yet, in fairness. This is also in-place.

Code like this makes my eyes cross :grinning:

src/chat/Eliza.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+class Eliza {
+public:
+    /**
+     * When loading the chat database file, this is used to map the name of a mob
+     * to the index number of a database that contains the keywords that it will
+     * recognise and the list of responses it can use.
+     */
+    struct NamedDatabase {
+        char *name{};
```
mattgodbolt, line 23: I think it'd be simple now to make this a std::string and get rid of the nasty. is the database name really mutable?

src/chat/KeywordResponses.cpp
mattgodbolt: right, and std::string

src/chat/KeywordResponses.hpp
mattgodbolt: cripes! :(

src/chat/KeywordResponses.hpp
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A response message with an assigned weight that's used when
+ * determining a random response to a keyword from the user.
+ */
+struct WeightedResponse {
+    int weight_;
+    char *message_;
```
mattgodbolt, line 13: can this be `const`? (if not a std::string)

src/chat/KeywordResponses.hpp
```diff
@@ -0,0 +1,38 @@
...
+class KeywordResponses {
+public:
+    int add_keywords(char *);
+    char *get_keywords() { return keywords_; }
+    int add_response(int weight, char *response_message);
+    WeightedResponse &get_response(int num) { return responses_[num]; }
+    const char *get_random_response();
+
+private:
+    char *keywords_{};
```
mattgodbolt, line 32: could these be const?

src/chat/KeywordResponses.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * Represents an keyword rule that is matched against the message supplied
+ * by the player, and the collection of response messages that can be sent
+ * if a keyword match is found.  Each keyword rule can contain logical operators.
+ * e.g. (what is your name|who are you).
+ * Each instance of KeywordResponses belongs to a specific Database.
+ */
+class KeywordResponses {
+public:
+    int add_keywords(char *);
+    char *get_keywords() { return keywords_; }
```
mattgodbolt, line 26: Some of these functions can be made member-const:

```cpp
char *get_keywords() const { return keywords_; }
const WeightedResponse &get_response(int num) const ... // IFF we can return a non-mutable
const char *get_random_response() const;
```

Const-correctness is another C++ superpower. It's not perfect, but starting with it is easier than retrofitting.

Ideally all those can return `const`  too. In some cases both are needed, which can be useful sometimes (this is how e.g. copy on write can be implemented in C++, you know when a caller needs a mutable reference and clone yourself etc).

src/chat/Database.cpp
snellers: I will check.  It came from the original.

src/chat/Database.cpp
snellers: Yes it's OG, will clean.

src/chat/Eliza.cpp
snellers: Yes it is,  this is the hairiest stuff in the OG code. Hard to tweak it without everything breaking.

src/chat/Database.cpp
snellers: Yep I can remove the return.

src/chat/Database.hpp
snellers: add_record() doesn't need a return actually.

src/chat/Eliza.hpp
snellers: I have put it on my TODO list

src/chat/KeywordResponses.hpp
snellers: Not right now, but probably once the way they're stored by KeywordResponses they can be

src/chat/KeywordResponses.hpp
snellers: keywords_ can! with tweaks to match().  The others can't until other changes.

PR #117, opened 2020-08-09
Extract the CHAR_DATA fields from descriptors

mattgodbolt:
Some subleties here, mainly around trying to unify and fix up some weird uses of "original" vs " character". See what you think. A lot is purely mechanical, but I extracted "person()" to mean "the actual person behind this descriptor" and hopefully used it correctly.

### DIFF
--- a/src/Descriptor.hpp
+++ b/src/Descriptor.hpp
@@ -53,13 +53,13 @@ class Descriptor {
     uint16_t port_{};
     bool processing_command_{};
     DescriptorState state_{DescriptorState::GetName};
+    CHAR_DATA *character_{};
+    CHAR_DATA *original_{};
 
     [[nodiscard]] std::optional<std::string> pop_raw();
 
 public:
     Descriptor *next{};
-    CHAR_DATA *character{};
-    CHAR_DATA *original{};
 
     explicit Descriptor(uint32_t descriptor);
     ~Descriptor();
@@ -108,4 +108,15 @@ public:
 
     void processing_command(bool is_processing) noexcept { processing_command_ = is_processing; }
     [[nodiscard]] bool processing_command() const noexcept { return processing_command_; }
+
+    [[nodiscard]] CHAR_DATA *character() const noexcept { return character_; }
+    void character(CHAR_DATA *character) noexcept { character_ = character; }
+    [[nodiscard]] CHAR_DATA *original() const noexcept { return original_; }
+    // do_ prefix because switch is a C keyword
+    void do_switch(CHAR_DATA *victim);
+    // do_ prefix because return is a C keyword
+    void do_return();
+    // Return the real-life "person" player character behind this descriptor.
+    [[nodiscard]] CHAR_DATA *person() const noexcept { return original_ ? original_ : character_; }
+    [[nodiscard]] bool is_switched() const noexcept { return original_ != nullptr; }
 };
--- a/src/act_comm.cpp
+++ b/src/act_comm.cpp
@@ -98,9 +98,9 @@ void announce(const char *buf, CHAR_DATA *ch) {
     for (d = descriptor_list; d != nullptr; d = d->next) {
         CHAR_DATA *victim;
 
-        victim = d->original ? d->original : d->character;
+        victim = d->person();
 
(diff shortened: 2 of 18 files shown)


src/Descriptor.hpp
```diff
@@ -108,4 +108,15 @@ class Descriptor {
...
     [[nodiscard]] bool processing_command() const noexcept { return processing_command_; }
+
+    [[nodiscard]] CHAR_DATA *character() const noexcept { return character_; }
+    void character(CHAR_DATA *character) noexcept { character_ = character; }
+    [[nodiscard]] CHAR_DATA *original() const noexcept { return original_; }
+    // do_ prefix because switch is a C keyword
+    void do_switch(CHAR_DATA *victim);
+    // do_ prefix because return is a C keyword
+    void do_return();
+    // Return the real-life "person" player character behind this descriptor.
```
snellers, line 119: yeah makes sense. And in this case, it's returning full CHAR_DATA, not just the PC_DATA, which it contains if the CD is a person :)

src/act_comm.cpp
```diff
@@ -98,9 +98,9 @@ void announce(const char *buf, CHAR_DATA *ch) {
     for (d = descriptor_list; d != nullptr; d = d->next) {
         CHAR_DATA *victim;
 
-        victim = d->original ? d->original : d->character;
+        victim = d->person();
```
snellers, line 101: I had to check what announce() actually did :P

src/Descriptor.hpp
```diff
@@ -53,13 +53,13 @@ class Descriptor {
     uint16_t port_{};
     bool processing_command_{};
     DescriptorState state_{DescriptorState::GetName};
+    CHAR_DATA *character_{};
```
snellers, line 56: ok they're private now so that'll  force everyone to use the new accessors, cool

src/act_wiz.cpp
```diff
@@ -1767,10 +1767,7 @@ void do_switch(CHAR_DATA *ch, const char *argument) {
         return;
     }
 
-    ch->desc->character = victim;
-    ch->desc->original = ch;
-    victim->desc = ch->desc;
-    ch->desc = nullptr;
+    ch->desc->do_switch(victim);
     /* change communications to match */
     victim->comm = ch->comm;
```
snellers, line 1772: I'm wondering whether it's rolling these last 3 lines into do_switch().

src/Descriptor.hpp
mattgodbolt: I nicked the name from a commented line in a few places. I _think_ it's safe to assume `person()->pc_data` is non-null...I dunno :)

src/act_comm.cpp
mattgodbolt: Yeah! this has been a great fun trip around the code. I forgot what `yell` did too!

src/comm.cpp
```diff
@@ -1536,9 +1533,9 @@ void act_new(const char *format, CHAR_DATA *ch, const void *arg1, const void *ar
             Descriptor *d;
 
             for (d = descriptor_list; d != nullptr; d = d->next) {
-                if (d->character != nullptr && d->character->in_room != nullptr && d->character->in_room->vnum == 1222
-                    && IS_AWAKE(d->character))
-                    send_to_char(buf, d->character);
+                if (d->character() && d->character()->in_room != nullptr && d->character()->in_room->vnum == 1222
```
snellers, line 1536: we should probably bundle all these magic room number constants into one place with constexprs. I know there's a few of them.

src/act_wiz.cpp
mattgodbolt: Yeah I wondered about that. It's a lot less about the descriptor and more about the CHAR_DATA though. For now I left them, thinking the descriptor is (mostly) about the lower-level stuff....though I can be swayed.

There's a lot of stuff in CHAR_DATA that should probably just be in descriptor.

I realise now the victim's `comm` is left forever in whatever state the last switcher had.

src/comm.cpp
mattgodbolt: Bloody right.

PR #135, opened 2020-08-16
Weather and time changes

mattgodbolt:
Hi folks. Like Tolkien's work, this change grew in the telling...

This extracts weather and time, and also moves nearly all time processing to use STL (and the coming-to-c++20 date library).

It's a big change; sorry. I thought I'd PR it though as I'm away and don't know when I'll actually commit it. Also I'm interested in whether you think this is an improvement or not (specifically the time handling, which is a little...er...crap in C++).

Notably lacking are tests for TimeInfoData and WeatherInfo which could actually be tested now. There's still a couple of globals around in this space, but I've tried to ensure the main code in those classes doesn't use the globals. So..I guess I should write some tests...before I commit. Thoughts welcomed of course!


src/db.cpp
```diff
@@ -235,45 +235,9 @@ void boot_db() {
...
-        lmonth = lday / 35;
-        time_info.month = lmonth % 17;
-        time_info.year = lmonth / 17;
-
-        if (time_info.hour < 5)
-            weather_info.sunlight = SUN_DARK;
-        else if (time_info.hour < 6)
-            weather_info.sunlight = SUN_RISE;
-        else if (time_info.hour < 19)
-            weather_info.sunlight = SUN_LIGHT;
```
mattgodbolt, line 255: Interestingly this contradicts what happened in the update loop

src/update.cpp
```diff
@@ -411,115 +415,14 @@ void mobile_update() {
...
  */
 void weather_update() {
-    char buf[MAX_STRING_LENGTH];
-    Descriptor *d;
-    int diff;
-
-    buf[0] = '\0';
-
-    switch (++time_info.hour) {
-    case 5:
```
mattgodbolt, line 421: I've chosen to have the weather update use this cycle:

5am: light (twilight I guess?
6am: rise
7pm: set (dusk?)
8pm: night

mattgodbolt:
Updated to add some tests (hardly comprehensive, but caught a few interesting things, like the inconsistent code from before!) Happy to change whatever re: day break etc. It has no functional effect I think.

src/Descriptor.cpp
```diff
@@ -36,7 +37,7 @@ const char *short_name_of(DescriptorState state) {
     return "<UNK>";
 }
 
-Descriptor::Descriptor(uint32_t descriptor) : channel_(descriptor), login_time_(ctime(&current_time)) {}
+Descriptor::Descriptor(uint32_t descriptor) : channel_(descriptor), login_time_(current_time) {}
```
snellers, line 40: I hunted around to find how it was gettin current_time, I see it's a global right now ;)

src/TimeInfoData.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (++month_ < MonthsPerYear)
+        return;
+    month_ = 0;
+    year_++;
+}
+
+TimeInfoData::TimeInfoData(Time now) {
+    using namespace date;
+    using namespace std::chrono;
+    // The epoch here, August 11, 1990 1:05:15 AM, is derived from the 650336715 in the original source.
```
snellers, line 62: How odd. I wonder why they opted for that.

src/TimeInfoData.cpp
```diff
@@ -0,0 +1,81 @@
...
+    static constexpr auto DikuEpoch = sys_days(1990_y / August / 11_d) + 1h + 5min + 15s;
```
snellers, line 63: aah, using those fancy user-defined operators!

src/TimeInfoData.cpp
```diff
@@ -0,0 +1,81 @@
...
+    const auto lhour =
```
snellers, line 64: translating:  lhour = the current count of "game hours since the diku epoch". Where each game hour == 30 seconds in real time.

snellers: Might be worth a comment there.

src/WeatherData.cpp
```diff
@@ -0,0 +1,139 @@
...
+namespace {
+
+std::string_view to_string(WeatherData::Sky sky) {
+    switch (sky) {
+    case WeatherData::Sky::Cloudless: return "cloudless";
+    case WeatherData::Sky::Cloudy: return "cloudy";
+    case WeatherData::Sky::Raining: return "rainy";
+    case WeatherData::Sky::Lightning: return "lit by flashes of lightning";
+    }
+    return "not as expected";
```
snellers, line 19: haha, the standard TV weather person's excuse :)

src/WeatherData.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+using namespace fmt::literals;
+
+WeatherData weather_info;
+
+namespace {
+
+std::string_view to_string(WeatherData::Sky sky) {
+    switch (sky) {
+    case WeatherData::Sky::Cloudless: return "cloudless";
```
snellers, line 14: At some point we should add some crazy weather like snow, hail, tornados :)

src/WeatherData.cpp
```diff
@@ -0,0 +1,139 @@
...
+            return WeatherData::Sun::Set;
+    }
+    return WeatherData::Sun::Dark;
+}
+
+}
+
+int WeatherData::pressure_direction(const TimeInfoData &tid) const {
+    if (tid.is_summer())
+        return mmhg_ > 985 ? -2 : 2;
```
snellers, line 48: I'd never noticed the diku/merc/rom folks had modeled air pressure.

src/WeatherData.cpp
snellers: Another thing we could do is make weather affect gameplay e.g. raining might reduce your 'scan' range. Cold/wet conditions might cause you to experience reduced dexterity etc.

src/comm.cpp
```diff
@@ -421,42 +418,12 @@ void game_loop_unix(Fd control) {
...
-        }
-
-        gettimeofday(&last_time, nullptr);
-        current_time = (time_t)last_time.tv_sec;
+        // Synchronize to a clock.
+        using namespace std::chrono;
+        using namespace std::literals;
+        constexpr auto nanos_per_pulse = duration_cast<nanoseconds>(1s) / PULSE_PER_SECOND;
+        auto next_pulse = current_time + nanos_per_pulse;
+        std::this_thread::sleep_until(next_pulse);
```
snellers, line 426: much cleaner

src/nth.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+std::string_view impl::suffix_for(size_t value) {
+    if (value > 4 && value < 20)
+        return "th";
+
+    if (value % 10 == 1)
+        return "st";
+    if (value % 10 == 2)
+        return "nd";
+    if (value % 10 == 3)
```
snellers, line 11: I think this may be wrong for numbers like 113, which should be th, but it'll come out as 113rd. It probably needs to go into the check on line 4.

src/update.cpp
snellers: seems fair

src/Descriptor.cpp
mattgodbolt: Sadly so; yes. I think it might serve us well to start calling out globals more obviously. It's not clear here. Maybe a naming convention for now/

src/TimeInfoData.cpp
mattgodbolt: Yeah! It was a surprise. It doesn't affect anything much as best I can tell but I wonder if that's the time (`date+=%s`) when the code was written. Fun code spelunking :)

src/TimeInfoData.cpp
mattgodbolt: Yeah, showing off. These come from the Date library (soon to be in c++20). I'm not 100% sure what I feel about them...

src/TimeInfoData.cpp
mattgodbolt: Right. Can do. This was mostly transliterated from the original unmolested, but I've used enough `auto` things and other affordances I should explain what's going on.

src/WeatherData.cpp
mattgodbolt: Nice ideas! Absolutely!

src/WeatherData.cpp
mattgodbolt: Was news to me too!

src/comm.cpp
mattgodbolt: And hopefully still equivalent!

src/nth.cpp
mattgodbolt: This is code from the original (or near enough). But good point, let me write and fix a failing test!

src/TimeInfoData.cpp
mattgodbolt: Done.

snellers (approved):

Issue #132, opened 2020-08-13
Disconnect + is broken


mattgodbolt:
`disconnect +` should disconnect by channel ID, but if one leaves a hanging telnet session and tried to disconnect:

```
WIZNET: Info from doorman: 4 is 24-148*** [#11640733258023999169]
[  4  Name] (unknown)@24-148*** [#11640733258023999169]
[  0  Play] TheMoog@24-148*** [#11640733258023999169]
2 users
> disconnect +4
Couldn't find a matching descriptor.
> disconnect + 4
Couldn't find a matching descriptor.
> help disconnect
DISCONNECT
Syntax: disconnect   <name>
Syntax: disconnect + <name>

Disconnect breaks a player's link to the game, dumping them back to the telnet
prompt.  This is a non-lethal way of getting rid of spammers and annoying
players.

Disconnect + is used to get rid of players who are not actually in the mud
but are connected and listed on 'sockets'. Rather than searching through
the active players, it looks through those connected via the descriptor
list.
```

mattgodbolt:
Per #138 this is not so much broken as either wrongly documented, or needing to support both `+player` or `+ID` to disconnect by searching descriptors instead of the character list.

mattgodbolt:
The +<number> doesn't seem to work though, needed to disconnect pre-logged in descriptors.

PR #138, opened 2020-08-18
Create a DescriptorList

mattgodbolt:
* Currently a global accessible by`descriptors()`
* Has some convenience methods for filtering and iterating over descriptors.
* LOTS of rephrasing with _hopefully equivalent_ code.


src/act_wiz.cpp
```diff
@@ -237,9 +242,9 @@ void do_disconnect(CHAR_DATA *ch, const char *argument) {
 
     if (argument[0] == '+') {
         argument++;
-        for (d = descriptor_list; d; d = d->next) {
-            if (d->character() && !str_cmp(d->character()->name, argument)) {
-                close_socket(d);
+        for (auto &d : descriptors().all()) {
```
mattgodbolt, line 245: this is the source of the error in #132

mattgodbolt: (I don't plan on fixing in this PR though)

mattgodbolt:
This is ready!

Things that are gnarly about this PR:

* I hope I got everything right. There's subtleties in some of the code. The main rationale is that many of the null checks were unnecessary: iff a descriptor is `is_playing()` THEN it must have a `character()`. Some of the checks seems bonkers paranoid to the point of confusion.
* the `ranges` library is used here to make filters with common subsets of the descriptor list. I've used it a few times before but I'm no expert. It's coming in C++20
* I've also used a couple of the `ranges::` convenience functions that take a range (e.g. a whole container). Things like `ranges::distance(x)` is equivalent to `std::distance(x.begin(), x.end());`. SUPER helpful if e.g. `x` was a function call returning a temporary (as our views are).

Let me know what you think. This is a troublesome one to review I know as so much looks the same but may be subtly different.

Thanks!

src/CMakeLists.txt
```diff
@@ -56,8 +56,10 @@ add_library(xania_lib STATIC
         CommandSet.hpp
         Descriptor.cpp Descriptor.hpp
         string_utils.cpp string_utils.hpp string_utils_impl.hpp
-        info.hpp WrappedFd.cpp WrappedFd.hpp)
-target_link_libraries(xania_lib xania_common crypt chat CONAN_PKG::fmt)
+        info.hpp
+        WrappedFd.cpp WrappedFd.hpp
+        DescriptorList.hpp DescriptorList.cpp)
+target_link_libraries(xania_lib xania_common crypt chat CONAN_PKG::fmt CONAN_PKG::range-v3)
```
snellers, line 62:
```
Range library for C++14/17/20. This code is the basis of the range support in C++20.
```

src/act_comm.cpp
```diff
@@ -90,23 +91,13 @@ void do_delete(CHAR_DATA *ch, const char *argument) {
...
 
-    for (d = descriptor_list; d != nullptr; d = d->next) {
-        CHAR_DATA *victim;
-
-        victim = d->person();
-
-        if (d->is_playing() && d->character() != ch && victim && can_see(victim, ch)
-            && !IS_SET(victim->comm, COMM_NOANNOUNCE) && !IS_SET(victim->comm, COMM_QUIET)) {
+    for (auto &d : descriptors().all_who_can_see(ch)) {
+        auto *victim = d.person();
```
snellers, line 98: There's a 'victim' check on 106, do we need that?

src/act_info.cpp
```diff
@@ -1620,22 +1624,19 @@ void do_whois(CHAR_DATA *ch, const char *argument) {
...
-        char const *class_name;
-
-        if (!d->is_playing() || !can_see(ch, d->character()))
-            continue;
+    for (auto &d : descriptors().all_visible_to(ch)) {
 
-        auto *wch = d->person();
+        auto *wch = d.person();
 
+        // TODO: can or should this be part of all_visible_to?
```
snellers, line 1631: I think we'll have to be wary of having too much of those canned filters, lest it become hard to tell which one to pick :) I guess it's all in the name.

snellers: I guess if we had a good selection of single-step filters, we can chain them ourselves too.

src/act_info.cpp
```diff
@@ -2009,19 +1990,15 @@ void do_credits(CHAR_DATA *ch, const char *argument) {
...
     if (arg[0] == '\0') {
-        snprintf(buf, sizeof(buf), "|cYou are in %s\n\rPlayers near you:|w\n\r", ch->in_room->area->areaname);
-        send_to_char(buf, ch);
-        found = false;
-        for (d = descriptor_list; d; d = d->next) {
-            if (d->is_playing() && (victim = d->character()) != nullptr && !IS_NPC(victim) && victim != ch
-                && victim->in_room != nullptr && victim->in_room->area == ch->in_room->area && can_see(ch, victim)) {
+        send_to_char("|cYou are in {}\n\rPlayers near you:|w\n\r"_format(ch->in_room->area->areaname), ch);
+        auto found = false;
+        for (auto &d : descriptors().all_visible_to(ch)) {
```
snellers, line 1999: If we find that there's more scenarios where we need to only collate the ch who are in the same area, this can probably become a descriptor streamer in its own right.

src/act_wiz.cpp
```diff
@@ -3465,41 +3454,36 @@ void do_rset(CHAR_DATA *ch, const char *argument) {
...
+        } else if (get_trust(ch) == MAX_LEVEL) {
+            // log even connections that haven't logged in yet
+            // Level 100s only, mind
+            name = "(unknown)";
+        }
+        if (name) {
             count++;
-            bug_snprintf(buf + strlen(buf), sizeof(buf), "[%3d %5s] (unknown)@%s\n\r", d->channel(),
-                         short_name_of(d->state()), d->host().c_str());
+            buf += "[{:3} {:>5}] {}@{}\n\r"_format(d.channel(), short_name_of(d.state()), name, d.host().c_str());
```
snellers, line 3477: do you need the c_str() here?

src/DescriptorList.hpp
```diff
@@ -0,0 +1,54 @@
...
+        return descriptors_ | ranges::views::transform([](auto &pr) -> Descriptor & { return pr.second; });
+    }
+
+    // Return all descriptors for characters in the "playing" state.
+    [[nodiscard]] auto playing() const noexcept {
+        return all() | ranges::views::filter([](const Descriptor &d) { return d.is_playing(); });
+    }
+
+    // Return all descriptors for playing characters, skipping the given character.
+    [[nodiscard]] auto all_but(const CHAR_DATA *ch) const noexcept {
```
snellers, line 30: pedantic, but all_but could also be all_playing_but  to distinguish it from the vanilla alls.

src/clan.cpp
```diff
@@ -81,18 +84,18 @@ void do_clantalk(CHAR_DATA *ch, const char *argument) {
...
-            candoit = 1; /* Yeah we can do it! */
-    } /* for all descriptors */
-
-    if (!candoit) {
-        send_to_char("Your clan lacks the necessary broadcast nexus, causing your vain telepathy to\n\rbe lost upon "
-                     "the winds.\n\r",
+    auto playing = descriptors().playing();
+    if (ranges::find_if(playing,
+                        [&](const Descriptor &d) {
+                            const auto *vix = d.person();
```
snellers, line 90: The 'vix' nomenclature seems particular to clan.cpp.

snellers: side note: I don't know why we require a clan hero/leader to be online for the clan chat to work. Seems novel but could be quite irritating (not that it matters right now!)

snellers (approved):
Big change! Nice to have the filtering stuff in. LGTM, some minor comments.

src/act_comm.cpp
snellers: I guess not given your earlier comment.

mattgodbolt: As best I can tell: no. I put some rationale in the PR itself but:
* `all_who_can_see` only returns playing descriptors that can "see" `ch`
* if a descriptor is playing then they must have at least one character associated with them. This I'm 99.998% sure (I can't find how it could not be true but there's a ton of defensive checks like this).
* `person` always returns the real character associated with a descriptor (if a char is switched then it returns `original` else the `character`). So for a "playing" character it's guaranteed to be non-NULL (as is `character()`, `original()` will be null for unswitched folks).

Might be worth putting those checks into the accessors themselves to maintain the invariant, and `throw` if not (though that would shut the MUD down).

src/act_info.cpp
mattgodbolt: Right. Thoughts and ideas welcomed here. I don't have a strong opinion: where I saw the same kind of check used in more than one place I extracted a canned filter. With a good name we're probably ok. With chained filters one has to remember the right sequence oneself (like `for (auto &x : descs.all() | playing | not_same(ch) | visible_to(ch))` vs `for (auto &x : descs.visible_to(ch))`

Composition is super cool but I'd imagine we'd find ourselves repeating the same pipeline list multiple times and then we'd make a helper function to do the work for us.

Very much open to trying this as well though!

src/act_info.cpp
mattgodbolt: Right. This is a good argument for the chainable filters: I thought a couple of times about "same room as" and "same zone as", but some were "including ch" and some "excluding ch", which I couldn't phrase with my monolithic functions. Definitely worth a think to have `all_in_zone()` and `all_in_room()` with a `| excluding(ch)` going on ... hmm. thanks!

src/act_info.cpp
snellers: I suspect simple method names is probably cleanest (and more reusable).  At the extreme, you could end up with NxM different combinations though hehe.

src/act_wiz.cpp
mattgodbolt: No!!! yay! Thanks! I have a mental TODO to look for c_str() cases that came along during integration work...thanks for eagle-eyed spotting that!

src/clan.cpp
mattgodbolt: re: hero online: Yeah it annoyed me during testing. I didn't change existing nomenclature, but will do.

mattgodbolt:
Thanks fella! I think I'll have a go at phrasing in_same_room and in_same_zone in a follow-up PR.

snellers (approved):

PR #149, opened 2020-08-20
Add a "player()" function.

mattgodbolt:
Many of the newer routines handled switched mobs by checking multiple
conditions. This extracts a "player()" function on the CHAR_DATA,
which returns the CHAR_DATA of the "real player" behind a character,
be they the character themselves, a controlling switched IMM, or
nullptr for a genuine, uncontrolled-by-player-character NPC.

I think this is equivalent to the earlier code and reduces complexity.
I wonder if the command interpreter can subsume this at some point,
with commands tagged as "no, this has to occur on the actual PC, not
the switched mob they might be controlling", to save further, though
that might be unobvious.

### DIFF
--- a/src/act_info.cpp
+++ b/src/act_info.cpp
@@ -801,13 +801,8 @@ void do_prompt(CHAR_DATA *ch, const char *argument) {
 
     /* PCFN 24-05-97  Oh dear - it seems that you can't set prompt while switched
        into a MOB.  Let's change that.... */
-
-    if (IS_NPC(ch)) {
-        if (ch->desc->is_switched())
-            ch = ch->desc->original();
-        else
-            return;
-    }
+    if (ch = ch->player(); !ch)
+        return;
 
     if (str_cmp(argument, "off") == 0) {
         send_to_char("You will no longer see prompts.\n\r", ch);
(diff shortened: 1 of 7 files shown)


src/act_info.cpp
```diff
@@ -1475,12 +1470,8 @@ void do_time(CHAR_DATA *ch, const char *argument) {
                      time_info.describe(), secs_only(boot_time), secs_only(current_time)),
                  ch);
 
-    if (IS_NPC(ch)) {
-        if (ch->desc->is_switched())
-            ch = ch->desc->original();
-        else
-            return;
-    }
+    if (ch = ch->player(); !ch)
```
mattgodbolt, line 1473: To work through this example:

The code before was:

* If we're an NPC:
  * if we're an NPC then to get to this point we must have a controlling descriptor (as we assume no mobprog runs "prompt", though...in general it could). We look at the descriptor, and see if it's switched. In honesty it _must_ be, because there's no realistic way it could be a PC's descriptor. (this code was already modifed from earlier so a mechanical change may have caused this slightly odd tautology). We go ask the descriptor for the "original", i.e. the CHAR_DATA associated with the switcher.  If we're not switched...we bail. Again I don't know how this could occur.
* if we're not an NPC, we're good.
Either way, we have bailed if we're somehow an NPC without a controlling player (though we probably would have crashed on original line 1479 as ch->desc would be `nullptr`), or else we have either the original `ch` if they were a PC to begin with, or their controlling `original()` CHAR_DATA.

In the new code:

we ask the CHAR_DATA for the player behind the character. It does the desc nullity check, then gets the desc->`person()` which returns either the char itself (if unswitched) or the original (if switched). Either way we get the original controlling char_data. If the desc is null, then we have no controller at all, and so it returns null. This little statement points `ch` at the controlling player, or returns if there wasn't one.

Obviously this code is subtle, so I'd encourage you to trace my logic with a critical eye. Thanks!

philjessies (approved):
Wow, that's a hell of a nice cleanup.

mattgodbolt:
Thanks @philjessies !

PR #170, opened 2020-08-23
Handle prompting consistently.

mattgodbolt:
* Use decriptor writes only in the output routines, instead of funky reacharound back to chars
* Handle colourisation of prompts and battle status in the same way.
* Reinstate colour in the battle status.


snellers (approved):

mattgodbolt:
Thanks pal

Issue #35, opened 2020-07-24
Feature: revisit the identity of Midgaard's main guardian. 


snellers:
Its name & emotes should be more on a fantasy theme.

snellers:
Closing this for now. More enhancements can be added later.

PR #181, opened 2020-08-26
Initial Concordius (#35).

snellers:


area/midgaard.are
```diff
@@ -174,19 +174,27 @@ AFN ABCD 0 0
 stand stand female 4500
 0 0 medium 0
 #3011
-frank bruno~
-Frank Bruno~
-Frank Bruno stands here protecting Xania.
+Concordius~
```
snellers, line 177: @mattgodbolt This is a first pass at the new guardian. There are a couple more things I want to do like add new behavior when he's fighting, and add some life to the "pet" eagle, Aquila.

snellers: Not doing those things yet so that I don't end up with a huge PR.

area/midgaard.are
```diff
@@ -174,19 +174,27 @@ AFN ABCD 0 0
...
+
+Although he seems to be quite relaxed, his upright posture and alertness
+indicate that he is ready to respond at a moment's notice.
+As he speaks, his sonorous voice leaves you feeling enlightened.
 ~
-giant~
-BMTV CDFJZ 1000 0
-86 30 1d1+6999 1d1+999 4d10+30 punch
+human~
+BMTV CDFHJZ 1000 0
```
snellers, line 196: Stats very similar to the old one, except +Sanctuary.

src/spec_concordius.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "comm.hpp"
+#include "merc.h"
+#include "string_utils.hpp"
+#include <fmt/format.h>
+#include <string_view>
+#include <vector>
+
+using namespace fmt::literals;
+namespace {
+
```
snellers, line 19: He has two modes: stand guard & fight. 
In stand guard mode, there's a 1:10 chance (every 4 or 5 seconds) he will do some random emote. If that doens't hit, he may pick one player out and social them. Nothing too fancy yet.

area/midgaard.are
```diff
@@ -2744,7 +2776,50 @@ officer blade~
...
+A
+23 5
+#3389
+toga linen~
+a fine linen toga~
+A fine linen toga lies in a crumpled heap.
+~
+cloth~
+armor AG AD
+20 20 20 28 0
```
snellers, line 2806: +20 armour is lower than usual for a level 85 item, but the +28 magic is quite high, and "rainbow resists" for the other magic types makes this potentially quite a good item for some encounters.

area/midgaard.are
mattgodbolt: Cheers chap

area/midgaard.are
```diff
@@ -174,19 +174,27 @@ AFN ABCD 0 0
...
+Concordius~
+Concordius stands here keeping the peace.
 ~
-Frank Bruno is covered in muscle, he is ready to fight and die defending Xania
-If you are a killer or thief you should beware as he will destroy you.
+Concordius is a tall imposing figure with a weathered face that suggests great
+experience. 
 
-                     ...BEHAVE WELL OR FACE HIM...
+Word has it that Concordius is some kind of arbiter, or arbitrator,
```
mattgodbolt, line 184: Haha love this.

area/midgaard.are
mattgodbolt: Seems appropriate

area/midgaard.are
```diff
@@ -2744,7 +2776,50 @@ officer blade~
...
+shield sun~
+the Arbiter's shield~
+A shield embossed with the image of the sun lies here.~
+bronze~
+armor ABG AJ
+24 24 24 20 0
+86 18 3000 P
+E
+shield~
+This bronze shield has been embossed with the symbol of  sun. Its outer edge
```
mattgodbolt, line 2789: `of  sun` - missing `the`?

area/midgaard.are
mattgodbolt: :+1:

area/midgaard.are
```diff
@@ -5657,7 +5732,7 @@ The concourse heads south, passing by a beautiful park.
 S
 #3271
 On the Concourse near the park~
-You are walking down the Concoruse, along the western wall of Midgaard.  To
+You are walking down the Concourse, along the western wall of Midgaard.  To
```
mattgodbolt, line 5735: hahah nice spot. in other projects I've integrated a spelling checker...probably worth doing something similar here :)

area/midgaard.are
```diff
@@ -6618,6 +6696,8 @@ M  3069 spec_guard
 M  3100 spec_cast_cleric
 M  3120 spec_cast_cleric
 M  3143 spec_mayor
+M  3011 spec_concordius
```
mattgodbolt, line 6699: Much more tractable :)

area/social.are
```diff
@@ -506,6 +506,16 @@ Might be a bit difficult.
 You ruffle your hair, wondering how far you can go before the rest think you're crazy.
 $n ruffles $s hair - weirdo!
 
+salute 0 0
```
mattgodbolt, line 509: Yay

src/spec_concordius.cpp
```diff
@@ -0,0 +1,85 @@
...
+static inline constexpr auto SOCIAL_DICE = 30u;
+static inline constexpr auto NUM_SOCIAL_ACTIONS = 7u;
+
+inline static const std::vector<std::string> emotes{
```
mattgodbolt, line 23: nice! Super minor: This can just be:

```
std::array emotes{"...", "..."};
```

(I learned the types are deduced in this case recently...) these are all constant strings, so `std::string` is overkill too, as written above this would be a std::array of const char*, but drop a `"$n tightens..."sv` `sv` suffix (with `using namespace std::literals`) to get an array of string_views, though either work. As does this in its current form!

src/spec_concordius.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+bool concordius_stands_guard(CHAR_DATA *ch) {
+
+    uint random = number_range(0, 100);
+    if (random > 90) {
+        act(emotes[random % emotes.size()], ch, nullptr, nullptr, To::Room);
+        return true;
+    }
+    bool found_someone = false;
+    for (auto &victim : descriptors().playing() | DescriptorFilter::same_room(*ch) | DescriptorFilter::to_character()) {
```
mattgodbolt, line 43: (Aside: one can loop over `ch->in_room_people` too; though includes npcs too. I'm still developing an understanding of when one iteration is better than another)

src/spec_concordius.cpp
```diff
@@ -0,0 +1,85 @@
...
+        default:
+        case 0: emote = "bow {}"_format(victim.name); break;
+        case 1: emote = "eye {}"_format(victim.name); break;
+        case 2: emote = "salute {}"_format(victim.name); break;
+        case 3: emote = victim.is_good() ? "punch {}"_format(victim.name) : "judge {}"_format(victim.name); break;
+        case 4: emote = victim.is_good() ? "highfive {}"_format(victim.name) : "peer {}"_format(victim.name); break;
+        case 5: emote = "smile {}"_format(victim.name); break;
+        case 6: emote = victim.is_good() ? "cheer {}"_format(victim.name) : "bonk {}"_format(victim.name); break;
+        }
+        interpret(ch, emote.c_str());
```
mattgodbolt, line 59: Your point about stringifying `interpret` is well-made, sir :)

src/spec_concordius.cpp
```diff
@@ -0,0 +1,85 @@
...
+    bool found_someone = false;
+    for (auto &victim : descriptors().playing() | DescriptorFilter::same_room(*ch) | DescriptorFilter::to_character()) {
+        uint random = number_range(0, SOCIAL_DICE);
+        if (random < SOCIAL_DICE - NUM_SOCIAL_ACTIONS)
+            continue;
+        found_someone = true;
+        std::string emote{};
+        switch (random % NUM_SOCIAL_ACTIONS) {
+        default:
+        case 0: emote = "bow {}"_format(victim.name); break;
```
mattgodbolt, line 51: overkill perhaps but maybe to avoid the spooky-action-at-a-distance of `NUM_SOCIAL_ACTIONS` and this list:

```cpp

struct ConcordiusPersonalEmote {
  std::string_view on_good;
  std::string_view otherwise;
};
std::array personal_emotes = {
ConcordiusPersonalEmote{"eye", "eye"},
ConcordiusPersonalEmote{"salute", "salute"},
ConcordiusPersonalEmote{"punch", "judge"},
//..etc
};

//...
auto &pers_emote = personal_emotes[random % personal_emotes.size()];
interpret(ch, "{} {}"_format(victim.is_good() ? pers_emote.on_good : pers_emote.otherwise, victim.name).c_str());
```

(untested! obviously ... and ironically the std:;array thing is annoying here, a vector <might> be better after all!) just a thought anyway

src/spec_concordius.cpp
snellers: done

src/spec_concordius.cpp
snellers: Yeah I had thought about that, but was trying to use the fancy new stuff. But I think I will swap to in_room as it is slightly more efficient. Just more old fashioned, that is until we make it a rangeable thing.

src/spec_concordius.cpp
snellers: Yeah funnily enough my first thought was to use an 'enum' but that was Java thinking. But I don't really like switch statements like this, I will try to improve it.

snellers: One of the interesting side effects of how things are set up is that the more players there are in the room, the more social he will seem.

PR #186, opened 2020-08-29
All CHAR_DATA names are now strings

mattgodbolt:
[not expecting a review, this is me staging this change for my own diffing/testing]

### DIFF
--- a/src/act_obj.cpp
+++ b/src/act_obj.cpp
@@ -38,24 +38,22 @@ void explode_bomb(OBJ_DATA *bomb, CHAR_DATA *ch, CHAR_DATA *thrower);
 
 /* RT part of the corpse looting code */
 
-bool can_loot(CHAR_DATA *ch, OBJ_DATA *obj) {
-    CHAR_DATA *owner, *wch;
-
+bool can_loot(const CHAR_DATA *ch, const OBJ_DATA *obj) {
     if (ch->is_immortal())
         return true;
 
     if (!obj->owner || obj->owner == nullptr)
         return true;
 
-    owner = nullptr;
-    for (wch = char_list; wch != nullptr; wch = wch->next)
-        if (!str_cmp(wch->name, obj->owner))
+    const CHAR_DATA *owner = nullptr;
+    for (const auto *wch = char_list; wch != nullptr; wch = wch->next)
+        if (matches(wch->name, obj->owner))
             owner = wch;
 
     if (owner == nullptr)
         return true;
 
-    if (!str_cmp(ch->name, owner->name))
+    if (matches(ch->name, owner->name))
         return true;
 
     if (owner->is_pc() && IS_SET(owner->act, PLR_CANLOOT))
@@ -1353,7 +1351,7 @@ void do_sacrifice(CHAR_DATA *ch, const char *argument) {
 
     one_argument(argument, arg);
 
-    if (arg[0] == '\0' || !str_cmp(arg, ch->name)) {
+    if (arg[0] == '\0' || matches(arg, ch->name)) {
         snprintf(buf, sizeof(buf), "$n offers $mself to %s, who graciously declines.", deity_name);
         act(buf, ch);
         snprintf(buf, sizeof(buf), "%s appreciates your offer and may accept it later.\n\r", deity_name);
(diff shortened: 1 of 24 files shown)


src/act_obj.cpp
```diff
@@ -2313,7 +2308,7 @@ void do_hailcorpse(CHAR_DATA *ch, const char *argument) {
             continue;
 
         for (current_obj = current_place->contents; current_obj != nullptr; current_obj = current_obj->next_content) {
-            if ((current_obj->item_type == ITEM_CORPSE_PC) && strstr(current_obj->short_descr, ch->name) != nullptr) {
+            if ((current_obj->item_type == ITEM_CORPSE_PC) && matches_inside(ch->name, current_obj->short_descr)) {
```
snellers, line 2311: This is actually interesting, it's a very similar problem to what I faced when working on the summoner - PC corpses don't include the player name in the 'name' field. I made the summoning routine match on short_descr just as this done.

src/act_wiz.cpp
```diff
@@ -1278,23 +1272,14 @@ void do_mstat(CHAR_DATA *ch, const char *argument) {
...
-                     race_table[victim->race].name, victim->sex == 1 ? "male" : victim->sex == 2 ? "female" : "neutral",
-                     victim->in_room == nullptr ? 0 : victim->in_room->vnum)
-            < 0)
-            bug("Buffer too small at "
-                "_file_name_"
-                ":"
-                "1296"
-                " - message was truncated");
-    } while (0);
+    bug_snprintf(buf, sizeof(buf), "Vnum: %d  Format: %s  Race: %s  Sex: %s  Room: %d\n\r",
```
snellers, line 1279: eventually bug_snprintf() itself can be ditched I suspect

src/challeng.cpp
```diff
@@ -426,9 +411,24 @@ void do_duel(CHAR_DATA *ch, const char *argument) {
     send_to_char("|cYou did not challenge that person.|w\n\r", ch);
 }
 
+namespace {
+void tell_results_to(int who) {
+    if (who == 0) {
```
snellers, line 416: I know you only moved it, but the who 0/1 thing is non-obvious.  We can redo it separately.

src/challeng.cpp
```diff
@@ -565,34 +540,32 @@ int fighting_duel(CHAR_DATA *ch, CHAR_DATA *victim) {
...
 /* Checks to see that various things are correct.  Only input is
    toggle to say what to do on error.  0=nothing, 1=abort duel. */
 static int check_duel_status(int f) {
-    int flag = 0;
+    unsigned int flag = 0;
 
     if (challenger != nullptr) {
-        if (strncmp(challenger->name, challenger_name, NAME_SIZE))
-            flag = 1;
+        if (challenger->name == challenger_name)
```
snellers, line 557: is the logic on these == right?  `if(strncmp` will only return 0 if it's equal?

src/mob_prog.cpp
```diff
@@ -1023,14 +1023,14 @@ void mprog_translate(char ch, char *t, CHAR_DATA *mob, const CHAR_DATA *actor, c
 
     *t = '\0';
     switch (ch) {
-    case 'i': one_argument(mob->name, t); break;
+    case 'i': one_argument(mob->name.c_str(), t); break;
```
snellers, line 1026: haha, minimal work on mob_prog.cpp I see, fair doos, that is another rat's nest and I'm pretty skeptical about it right now.

snellers (approved):
Looks  like a pretty mega step forward. One or two minor comments.

src/act_wiz.cpp
mattgodbolt: Absolutely!

src/challeng.cpp
mattgodbolt: Totally: an `enum class` or similar...or anything...

src/mob_prog.cpp
mattgodbolt: 100% I suspect it's not long for this world so I didn't spend too much time making it better!

src/challeng.cpp
mattgodbolt: Wow! Great spot: I was 100% wrong here!!!

src/mob_prog.cpp
snellers: Do I get a free spell or a customized item? :)

src/challeng.cpp
mattgodbolt: Fixed! Am going to go over everything one more time then I'll merge!

Issue #182, opened 2020-08-27
Issues with iterating over lists that are mutated as we go (specifically pets)


mattgodbolt:
There's a lot of code that loops over CHAR_DATAs in various containers with code like:

```cpp
for (auto *ch = blah; ch; ch = ch->next) {...}
```

 Some code takes defensive action "what if the the `ch` is removed/deleted/moved...

```cpp
CHAR_DATA *ch_next;
for (auto *ch = blah; ch; ch = ch_next) {
 ch_next = ch->next;
 ...maybe do stuff that deletes `ch` or moves it out of the room or whatever...
}
```

This is "ok" but:

* Users have to remember which pattern to use, which is "safe" vs "I'm sure nothing will go away"...
  - probably a general iteration solution can make this easier
* Some code paths remove _more than one_ char.

In the first case, some smart iteration will make this "go away" (mostly).

In the second case, smart iteration may be able to be taught (I have ideas)...

The second case can probably happen when:
* CHAR_DATA list is `TheMoog -> TheMoog's pet -> nullptr`
* MUD is shutdown.
* [This code runs](https://github.com/mattgodbolt/xania/blob/850383fed3d31c60ece32907fe906ab29c863f6e/src/update.cpp#L667-L675)
```cpp
    for (ch = char_list; ch != nullptr; ch = ch_next) {
        ch_next = ch->next;

        if (ch->desc != nullptr && ch->desc->channel() % save_every_n == save_number)
            save_char_obj(ch);

        if (ch == ch_quit)
            do_quit(ch, "");
    }
```
* in `do_quit` the character _and their pet_ are removed
* the `ch_next` is invalid (it will point at the pet that's been removed)
* :fire:

PR #199, opened 2020-08-31
Initial work on a vector of "safe to remove" things

mattgodbolt:
Tested, and working. But I wonder if this is the right approach:

* a vector is cool but maybe a list is great after all?
* supports arbitrary nesting of safe_iteration, which seems overkill
* supports removal of items other than the "currently iterated over"
  (this is untested as the old code path didn't support it).
* doesn't support "clear" while safe iterating
* "safe iteration" is a modification_safe_for_each, not general-purpose
  iteration. Should _probably_ support any kind of iteration.

I'm thinking this pattern is common enough that we should come up with a
decent solution that "just works" in all cases. I would like to be able
to write:

```cpp
ranges::for_each(container, [](auto &elem) { if (elem.bad()) container.remove(elem); }
```

and have it "Just Work". Maybe that's overkill, it's certainly atypical C++ code.

The old code is a bit split-brained about whether it used `it = it->next` or `it = it_next`,
and the old code also didn't support multiple removals during iteration (see
for example #182).

### DIFF
--- a/src/AffectList.cpp
+++ b/src/AffectList.cpp
@@ -5,32 +5,21 @@
 #include <range/v3/iterator/operations.hpp>
 
 AFFECT_DATA &AffectList::add(const AFFECT_DATA &aff) {
-    auto new_aff = new AFFECT_DATA{aff};
-    new_aff->next = first_;
-    first_ = new_aff;
-    return *new_aff;
+    data_.insert(data_.begin(), aff);
+    return data_.front();
 }
 
 void AffectList::remove(const AFFECT_DATA &aff) {
-    if (&aff == first_) {
-        first_ = aff.next;
-    } else {
-        AFFECT_DATA *prev;
-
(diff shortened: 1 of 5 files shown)


src/AffectList.cpp
```diff
@@ -5,32 +5,21 @@
...
-                break;
-            }
-        }
-
-        if (prev == nullptr) {
-            bug("Affect_remove: cannot find paf.");
-            // TODO may leak
-            return;
-        }
+    auto index = &aff - data_.data();
```
snellers, line 13: that's interesting, but I guess it works because vector always has things in a consistent order? Or is std::distance a "safer" way to do it?

mattgodbolt: `std::distance` works on iterators, not references. I'm sure there's a better way to do this: this totally relies on the vector being contiguous.

mattgodbolt:
Closing in favour of #233

PR #210, opened 2020-09-05
One way of fixing the leak in MemFile...

snellers:


mattgodbolt (approved):
LGTM! It was a cute trick to use unique_ptr to do the hard work, but as we need to call `::free` we might as well do it all ourselves! Thanks.

Can you remove the `#include <memory>` now it's unused! Cheers for fixing!!

PR #220, opened 2020-09-13
MobIndexData changes

mattgodbolt:
* Made a few things ClassCase vs SHOUTY_OLD_STYLE
* Extracted a "Dice" class. Wrote tests.
* Extracted a few macros, get_hitroll and get_damroll
* Made test output a bit nicer (CatchFormatters.hpp)


mattgodbolt:
Just here for show & tell. Will merge tomorrow if all still looks good (tested a bit locally, will test more).

mattgodbolt:
Testing looks good: will fix up build errors.

src/MobIndexData.hpp
```diff
@@ -0,0 +1,57 @@
...
+    long vuln_flags{};
+    sh_int start_pos{};
+    sh_int default_pos{};
+    sh_int sex{};
+    sh_int race{};
+    long gold{};
+    long form{};
+    long parts{};
+    sh_int size{};
+    sh_int material{}; // TODO: is this actually used in any meaningful way?
```
snellers, line 49: I'm fairly sure materials are used in things like calculations for breath attack, and the material vulnerabilities. Whether they look at this field specifically though I don't know without checking.

src/act_wiz.cpp
```diff
@@ -1358,25 +1356,14 @@ void do_mfind(Char *ch, const char *argument) {
...
-    for (int vnum = 0; nMatch < top_mob_index; vnum++) {
-        if (auto *pMobIndex = get_mob_index(vnum)) {
-            nMatch++;
-            if (is_name(argument, pMobIndex->player_name)) {
-                found = true;
-                buffer += fmt::format("[{:5}] {}\n\r", pMobIndex->vnum, pMobIndex->short_descr);
-            }
-        }
-    }
+    auto buffer = ranges::accumulate(all_mob_indexes() | ranges::views::filter([&](const auto &mob) {
```
snellers, line 1359: nice

src/MobIndexData.hpp
mattgodbolt: I couldn't see the usage but I was also knee-deep in other stuff :)

src/act_wiz.cpp
mattgodbolt: Yeah, that was a bit of a flex :P

PR #233, opened 2020-09-30
New iteration-safe list

### LINKED ISSUE #182: Issues with iterating over lists that are mutated as we go (specifically pets)

### LINKED ISSUE #218: Crash seen in prod in affect_modify
Stack trace:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005571e81580f9 in affect_modify(Char*, AFFECT_DATA const&, bool) [clone .part.0] ()
(gdb) bt
#0  0x00005571e81580f9 in affect_modify(Char*, AFFECT_DATA const&, bool) [clone .part.0] ()
#1  0x00005571e81582f1 in affect_remove_obj(OBJ_DATA*, AFFECT_DATA const&) ()
#2  0x00005571e819c4ab in obj_update() ()
#3  0x00005571e819c7c1 in update_handler() ()
#4  0x00005571e8135801 in game_loop_unix(Fd) ()
#5  0x00005571e812c311 in main ()
```

One character online but wasn't doing anything at the time (CC @pierrewery )

mattgodbolt:
Nearly there: some todos. Don't feel it needs a look just yet.

fixes #218
fixes #182
obviates #199


mattgodbolt:
Notably, this replaces one O(n) algorithm with another. 

Previously a remove had the "object pointer" and thus had all the info needed to remove the item (the object contains the "next"). But a linear search was needed to find "prev" to unlink it. O(n)

Now we have a doubly-linked list, but the prev/next items aren't inside the items (that is, the list is no longer _intrusive_). So an iteration is needed to find the list node that refers to the object. O(n) too.

In an ideal world we could use a boost intrusive_list style thing which would embed the prev/next in the object. the ref counting/tombstones would need to be changed completely though. I toyed with keeping a list of live iterators to update, but didn't go down this route. There are tradeoffs in both approaches and frankly I've thought too much about this already!

mattgodbolt:
@snellers -- don't feel you have to look at this, I'll probably merge tomorrow. But thought it worth you taking a quick look if you have a mo: mainly the core algorithm, and then just a random sample of the changes. I tested things like boats, keys, affects, loads and saves, but I am sure I broke at least something.

In general this means lists can be used like "normal" without ever worrying about things like "pNext" and such. Which is great.

src/GenericList.hpp
```diff
@@ -0,0 +1,203 @@
...
+public:
+    struct Elem {
+        OptType item;
+        mutable size_t count{};
+        [[nodiscard]] bool removed() const { return !item; }
+    };
+    using MutIter = Iter<T, GenericList<T>, typename std::list<Elem>::iterator>;
+    using ConstIter = Iter<const T, const GenericList<T>, typename std::list<Elem>::const_iterator>;
+
+    // Add to the back of the list.
```
snellers, line 32: This forced my brain to do a hard parse, but it made sense :)

src/GenericList.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+    // Construct a list of the given items.
+    template <typename... Args>
+    static GenericList of(Args &&... args) {
+        GenericList result;
+        (result.add_back(args), ...);
+        return result;
+    }
+
+    // Iteration support.s
```
snellers, line 69: stray .s

src/GenericList.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+    // Add to the back of the list.
+    T &add_back(T item) { return *list_.emplace_back(Elem{std::move(item), 0}).item; }
+    // Add to the front of the list.
+    T &add_front(T item) { return *list_.emplace_front(Elem{std::move(item), 0}).item; }
+
+    // Remove an item. Removes by comparison to T, so comparing Ts must be defined.
+    // Returns true if item was successfully removed.
+    bool remove(const T &item) {
+        if (auto it = ranges::find(list_, item, &Elem::item); it != list_.end()) {
```
snellers, line 40: I am struggling to find the definition of this flavour of find. The headers are unintelligble and I couldn't find a decent reference page.

src/magic.cpp
```diff
@@ -3296,12 +3296,25 @@ void spell_refresh(int sn, int level, Char *ch, void *vo) {
         act("$N looks invigorated.", ch, nullptr, victim, To::Char);
 }
 
+namespace {
+bool is_cursed(const OBJ_DATA *obj) { return IS_OBJ_STAT(obj, ITEM_NODROP) || IS_OBJ_STAT(obj, ITEM_NOREMOVE); }
```
snellers, line 3300: better, thanks

src/AffectList.hpp
```diff
@@ -1,48 +1,36 @@
...
 
 #include "AFFECT_DATA.hpp"
-#include "GenericListIter.hpp"
-
-#include <utility>
+#include "GenericList.hpp"
 
 class AffectList {
-    AFFECT_DATA *first_{};
+    GenericList<AFFECT_DATA> list_;
```
snellers, line 7: Now that we have GenericList,  do we need to have AffectList wrapper type any more? i.e. could we be using `GenericList<AFFECT_DATA>` everywhere?

src/Char.hpp
```diff
@@ -38,7 +38,7 @@ struct Char {
     Descriptor *desc{};
     AffectList affected;
     NOTE_DATA *pnote{};
-    OBJ_DATA *carrying{};
+    GenericList<OBJ_DATA *> carrying;
```
snellers, line 41: I noticed AffectList contains a GenericList of AFFECT_DATA,  while this one is a list of OBJ_DATA pointers.  And I'm wondering what the pattern ought to be going forward.

src/GenericList.hpp
mattgodbolt: Zoiks! Fixed, thanks :)

src/GenericList.hpp
mattgodbolt: Right! Yeah, this one's more tricky than it _ought_ to be. The documention is lamentable; I hope it improves as it's now part of the standard (C++20).

Anyway; in general these kinds of range functions take params of the form: `(range, thing_to_do_or_find, [comparison operation in some cases], projection)`. The last few params are optional. In the case of `find`, there's no `comparison operation`, it's always equals. The optional "projection" function takes each element and essentially says "what thing should I actually compare against", the default being "the whole thing". As a special case, the "function" can be a reference to a member variable, or a member function, in which case it's the same as writing a function that returns that variable.

All that to say, we want to find items in the list by their "item" member variable. So we project each `Elem` to its `item` to find by `item`.

The "pass a ref to the field" shortcut is used a bit in the rest of the PR and elsewhere, it's a great way when very commonly you want to find by a field, or sort by a field etc.

mattgodbolt: TLDR find "item" (in this case of type `T`) in the list of `Elem`s by looking at each `Elem`'s `item` field (which is a `T`)

src/AffectList.hpp
mattgodbolt: I wondered about this. There's a couple of useful helper functions on the AffectList that were "nice" to keep for now, mainly "find by skill". But I agree I think this could probably go. In fact, looking at `find_by_skill` : 

```cpp
    if (auto it = ranges::find(*this, skill_number, [](auto &af) { return af.type; }); it != end())
        return &*it;
    return nullptr;
```

this is an example of the code before I knew I could shortcut the lambda to "project" an affect to its skill. So now I could write this as:

```cpp
    if (auto it = ranges::find(*this, skill_number, &AFFECT_DATA::type); it != end())
        return &*it;
    return nullptr;
```

...at which point the function is "simple enough" to maybe inline where it's used. Maybe? Sometimes it's nice to have a human-friendly name for these operations. We can make free functions that take `GenericList<AFFECT_DATA>` and find by skill I guess, so callers would do:

```cpp
auto *af = find_by_skill(list_, skill_number);
```
in that case. More "FP" than OO :)

Either way; for now I'll leave this I think and we can always remove it later (which I'm pretty sure we should do on balance).

src/Char.hpp
mattgodbolt: Right! This is owning, vs-non-owning.

The AffectList _owns_ the objects. They're part of the list, and when the list is destroyed so are the AFFECT_DATAs inside.

The carrying is a _reference_ (in the non-c++ sense) to the carried object. In this case the owner is (sort of) the global `object_list`. At least, that's the thing responsible for tidying it up.

Right now the code hasn't been refactored to use C++ trickery to make the "object_list" actually "own" the data; `extract_obj` is responsible for removing it from all the lists and then ultimately `delete`-ing it.

I went back and forth on this a number of times. Trying to retrofit strict C++ style ownership is the goal but it was already a big change so I left `extract_obj` et al alone. `AffectList` had already been refactored to "own" its items so I changed that.

In the longer-ish term I can actually see us moving to a model where 99% of `extract_obj` goes into the OBJ_DATA destructor, and then extract_obj then becomes "remove this object from the object_list" (which would then "own" the OBJ_DATA either by it being a `GenericList<OBJ_DATA>` or something like `GenericList<unique_ptr<OBJ_DATA>>`.

I really this is really very discursive: I have been back and forth on this a dozen times: most of the time I spent on this whole PR was analysis paralysis on these kinds of decisions; so for now I've chosen a local minima to get something in that's an incremental improvement (I hope!)

snellers: makes sense

PR #234, opened 2020-10-01
Change signal handling when a debugger is attached.

### LINKED ISSUE #191: xania crashes if gdb connects after it starts then tries to set a breakpoint

mattgodbolt:
Continue to block SIGINT and handle it as a shutdown, but if a
debugger is detected, instead raise a SIGTRAP to hit a breakpoint.

Closes #191 (I think). I tested it with running under `gdb`, and with `gdb -p` and it did the Right Thing™ as far as I could tell, but I didn't test under vscode (sorry; I was lazy...)

Would you mind checking to see if this fixes things for you, @snellers? Thanks!

### DIFF
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -44,8 +44,10 @@
 #include <ctime>
 #include <fcntl.h>
 #include <fmt/format.h>
+#include <fstream>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <streambuf>
 #include <string>
 #include <sys/signalfd.h>
 #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


src/comm.cpp
```diff
@@ -283,6 +302,7 @@ void game_loop_unix(Fd control) {
          * Poll all active descriptors.
          */
         FD_ZERO(&in_set);
+        FD_ZERO(&in_set);
```
snellers, line 305: FD_ZERO on in_set twice?

src/comm.cpp
```diff
@@ -126,6 +128,21 @@ void greet(Descriptor &d) {
     d.write(greeting->text());
 }
 
+namespace {
+bool is_being_debugged() {
+    std::ifstream t("/proc/self/status");
```
snellers, line 133: aha yes, I'd read about this :)

snellers: There ought to be a standard API for this sort of thing like in Windows.

snellers:
@mattgodbolt  works perfectly, thanks!

src/comm.cpp
mattgodbolt: Yeah it's a pain in the proverbials...this is the best I could find.

src/comm.cpp
mattgodbolt: Oh...sometimes it doesn't work first time, and you have to do it again...
er....whoops no, just a typo :blush:

mattgodbolt:
Thanks chap!

PR #247, opened 2020-11-06
logrotate uses copytruncate mode as the stderr/stdout never get close…

### LINKED ISSUE #63: log rotation isn't working
We don't reopen our log files, so the log rotation set up in systemd isn't working. Probably should be fixed by a more sensible logging system.

snellers:
…d, and only rotate if the log file gets large enough. Resolves #63

### DIFF
--- a/infra/scripts/xania.logrotate
+++ b/infra/scripts/xania.logrotate
@@ -2,5 +2,7 @@
 /home/xania/data/log/doorman.log
 {
     daily
+    size 1M
+    copytruncate
     rotate 60
 }


infra/scripts/xania.logrotate
```diff
@@ -2,5 +2,7 @@
 /home/xania/data/log/doorman.log
 {
     daily
+    size 1M
```
snellers, line 5: @mattgodbolt  should be a simple fix for the 'zero byte log file' issue. It'll also cut down on redundant rotation where log files are really small. 

As a side note, the `append:` mode we set in the service configs for `StandardOutput` only works with systemd >= 240, and by default 18.04 only ships with 237, which only supports `file:`. Discovered while trying this stuff out locally.

snellers: To deploy this I'll end up copying this file into the right place, what the init scripts do.

mattgodbolt: Sounds perfect.

mattgodbolt (approved):

Issue #221, opened 2020-09-14
Persistent directory for notes.txt, bugs.txt and ban.lst


mattgodbolt:

mattgodbolt:
See also #53

PR #252, opened 2020-12-31
Revamp and C++ify notes

mattgodbolt:
* All C++ all the time
* Mostly global-free. Only one singleton bridging to Xania
* Tests!

Still writes to area/notes.txt which means this does not help with #221,
but at least we can test the fix as and when.


mattgodbolt:
Some notes:
* times are now formatted as `gmtime` else my tests fail (they were in "local" time where local is Chicago for me...)
* some minor tweaks to a few other routines

This is _so low priority_ so don't fret. If I don't hear in a week or so I might merge anyway unless you have a strong feeling you want to look. This was more to get me back into Xania and C++ :)

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+#include <range/v3/algorithm/find_if.hpp>
+#include <range/v3/view/filter.hpp>
+
+#include <cctype>
+#include <cstdio>
+#include <functional>
+
+static const char *NOTE_FILE = "notes.txt";
+
+bool Note::is_to(const Char &ch) const {
```
snellers, line 27: Not that it matters today, but one idea I had is being able to send a note to your whole clan.  Perhaps we'll do that some day if there are players and clans!

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+    if (is_name(ch.name, to_list_)) {
+        return true;
+    }
+    return false;
+}
+
+bool Note::sent_by(const Char &ch) const { return matches(ch.name, sender_); }
+
+void Note::save(FILE *file) const {
+    fmt::print(file, "Sender {}~\n", sender_);
```
snellers, line 46: fmt::print, very handy :)

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+    fmt::print(file, "Subject {}~\n", subject_);
+    fmt::print(file, "Text\n{}~\n", text_);
+}
+
+void Notes::erase(const Note &note) {
+    auto it = ranges::find_if(notes_, [&](const Note &data) { return &data == &note; });
+    if (it != notes_.end())
+        notes_.erase(it);
+}
+Note &Notes::add(Note note) { return notes_.emplace_back(std::move(note)); }
```
snellers, line 59: Q: The arg is not passing a reference or a pointer here,  which makes me think it would be a copy by value, but would passing as a reference be any better?

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+int Notes::num_unread(const Char &ch) const {
+    int notes = 0;
+    for (auto &note : notes_) {
+        if (note.is_to(ch) && !note.sent_by(ch) && note.date_stamp() > ch.last_note) {
```
snellers, line 64: Q: it doesn't matter, but might there be a `ranges` way to iterate & filter the notes? (Just thinking in terms of how it can be done with Java streams).

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+            notes++;
+        }
+    }
+    return notes;
+}
+
+Note *Notes::lookup(int index, const Char &ch) {
```
snellers, line 71: I notice the first args of these two lookup() aren't const

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+}
+
+static Note &ensure_note(Char &ch) {
+    if (!ch.pnote)
+        ch.pnote = std::make_unique<Note>(ch.name);
+    return *ch.pnote;
+}
+
+Note Note::from_file(FILE *fp) {
+    auto expect = [&](std::string_view expected) {
```
snellers, line 109: I like the nested lambdas

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+    sub_commands.add("+", &NoteHandler::add_line);
+    sub_commands.add("-", &NoteHandler::remove_line);
+    sub_commands.add("subject", &NoteHandler::subject);
+    sub_commands.add("to", &NoteHandler::to);
+    sub_commands.add("clear", &NoteHandler::clear);
+    sub_commands.add("show", &NoteHandler::show);
+    sub_commands.add("post", &NoteHandler::post);
+    sub_commands.add("send", &NoteHandler::post);
+    sub_commands.add("remove", &NoteHandler::remove);
+    sub_commands.add("delete", &NoteHandler::delet, MAX_LEVEL - 2);
```
snellers, line 181: hah `delet`

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+            show_note(*num, *note);
+        } else {
+            ch.send_line("You have no unread notes.");
+        }
+    } else {
+        ch.send_line("Note read which number?");
+    }
+}
+
+void NoteHandler::list(Char &ch, [[maybe_unused]] ArgParser) {
```
snellers, line 242: cute, I guess it's nicer than void casting

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+    if (!ch.pnote || ch.pnote->text().empty()) {
+        ch.send_line("There is no text to delete.");
+        return;
+    }
+    ch.pnote->remove_line();
+    ch.send_line("Ok.");
+}
+
+void NoteHandler::subject(Char &ch, ArgParser args) {
+    ensure_note(ch).subject(args.remaining());
```
snellers, line 295: Hmm,  if someone doesn't set the subject, or uses `note subject <empty>` do we save/send a note without one?

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+    ch.send_line("To: {}|w", note.to_list().empty() ? "(No recipients)" : note.to_list());
+    if (note.text().empty())
+        ch.send_line("(No message body)");
+    else
+        ch.send_to(note.text());
+
+    ch.send_line("|w");
+}
+
+void NoteHandler::remove(Char &ch, ArgParser args) {
```
snellers, line 325: It occurred to me that over a long time you could accrue a tons of notes in the file, and having something to sweep away old ones could be useful (eventually).

src/Note.cpp
```diff
@@ -0,0 +1,378 @@
...
+    note.text_ = expect_str("text");
+    return note;
+}
+
+void Note::subject(std::string_view subject) { subject_ = subject; }
+void Note::date(Time point) {
+    date_ = formatted_time(point);
+    date_stamp_ = point;
+}
+void Note::to_list(std::string_view to_list) { to_list_ = to_list; }
```
snellers, line 133: I'd been wondering if there ought to be some sort of limit on how many recipients you can specify. We have the regular max input length, I guess that covers it.  A bonus thought,  the to list (from ArgParser) could be uniq'd too :)

src/test/NoteTest.cpp
```diff
@@ -0,0 +1,313 @@
...
+                CHECK(index == 1);
+            }
+            SECTION("no more") { CHECK(!notes.lookup(Time(120s), jenny, &index)); }
+        }
+    }
+}
+
+TEST_CASE("Note command handling") {
+    // Pretty nasty way of hacking out something that looks enough like a PC to write tests.
+    // Ideally we'd extract a lot more interfaces and test using mocks.
```
snellers, line 194: +1!

snellers:
Looks nice!

snellers (approved):

src/Note.cpp
mattgodbolt: TLDR; this nets out as more efficient in some cases if you know you're going to be copying one of your parameters. The clang linter recommends this approach in many cases (so I've gotten used to doing it as a habit).

---

This is a fun one, and mainly goes by the theory that if you're going to be taking of copy of something (as we are when we put the note into the vector), then you should make that copy explicit. In this instance the "copy" happens in the parameter name.

The interesting thing about this is that by _moving_ from the parameter you can get nearly perfect forwarding/construction behaviour.

Take for example some code like this, which takes a copy of a string:

```
using namespace std; // lazy
struct Thing {
  string s_;
  void set_s(string s) { s_ = move(s); }
};
```

Now consider what happens when we do something like `set_s("moo");`:

* the `const char*` gets converted into a temporary `string` via `string`'s constructor that takes a const char*. That does allocations^ etc and copies the "moo" into the new temporary `string` object`
* that temporary is used to construct the parameter value `s` in the `set_s`. The argument to the constructor of this `s` is a temporary - so it's of type `string &&`. The `string&&` constructor can "steal" the contents of the temporary `string&&` which makes it super cheap (no allocations, just swapping pointers)
* the `s_ = std::move()` runs. The `s_`'s `operator=(string &&)` is called (as the `move` turns the `string` into an r-value). This _also_ does the steal-the-innards trick.
* All the destructors for the temporaries (the original "moo" object and the pass-by-copy param `s`) run. They have no work to do as they've been "stolen" from.

Short version: exactly one "real" string construction is done, and then there's a bit of pointer movement, but nothing more expensive.

If, however, we change to a `const string &` in `set_s`:
* the `const char *` gets converted into a temporary `string` as before.
* it is passed by reference (super cheap) to `set_s()` - no work done here at all.
* In the `s_ =  s;` we copy the string object. This is another allocation and copy, as we can't "steal" anything from anyone
* we then destroy the temporary `string`

This means we do the allocation twice, copy the string data twice, and do a deallocation once. More expensive! https://godbolt.org/z/n9vqx9

---

^ ok so actually no real allocations are done in this case as there's something called the "small string optimization" (SSO) kicking in; the string object holds small strings _by value_ inside the object. But,,,the work is still there. If you change the string to something longer you see the work https://godbolt.org/z/n9vqx9 , but it's more complex :) The expense is in the `_M_create` method you'll see in the asm output.

src/Note.cpp
mattgodbolt: Definitely! Good point, then I can use `count`. Will fix

src/Note.cpp
mattgodbolt: Right; in general I've not made value-types const. It doesn't change the type signature at all, that is `void foo(int);` and `void foo(const int);` are identical in the type system, so I tend not to do it as it can be confusing. I know some people make it `const int` so that  _variable_ of the parameter is `const` within the body, which can be useful sometimes, but I haven't gotten into the habit of that.

mattgodbolt: (in fact in the first case, the `index` _is_ changed, so couldn't be `const` without changing the code (which might be clearer, but this was the OG code).

src/Note.cpp
mattgodbolt: yeah sucks, right? The worst one to hit is `register` as we've not really used that as a keyword in anger since the early 90s

src/Note.cpp
mattgodbolt: No, the `post` code has an error for that (and there's even a test! shock)

src/Note.cpp
mattgodbolt: Notes older than 2w are deleted on load.

src/Note.cpp
mattgodbolt: Right :)

mattgodbolt:
Well, _one_ of the test hosts died with:

```
[-325297776] TheMoog: So here it is
..."
 ==
"
[  0] TheMoog: So here it is
...
```

so...time to work out what travesty I've managed to slip past the regular tests...

I suspect I know actually, and it's something I meant to change! Watch this spaacespace.

mattgodbolt:
Thanks for the comprehensive review btw! :)

src/Note.cpp
snellers: Aha!

PR #255, opened 2021-05-09
pfu player file upgrade utility and prepare-pfiles.sh upgrade script

snellers:
A new tool called `pfu` plus a pair of wrapper scripts. This enables us to proactively upgrade all player files as part of a release. It works by loading the game world, simulating logging in each char and then executing a list of tasks on the char before saving it. 

A grep reveals that on the host, all existing pfiles are either v3 or v4.

**Explicit Upgrades**

- Right now there's only one task - one that emulates  `reset_char()`  which previously was always run on login even if the Char didn't need it, and this is largely about resetting & recalculating any attributes that can be modified by item & spell effects.
- All pfiles are being upgrade to v5.

**Implicit Upgrades**

As well as the explicit tasks, there are several updates that pfu will make to existing pfiles simply by saving them:

- Unique Objects: Any objects a char owns that has been flagged UNIQUE gets its "ExtF" bits updated to set that bit.
- Adds empty pronouns.
- Removes `Info_` finger fields that we ditched last year.
- Removes `NewsRead` field.
- Some chars got a new bit in their `ExtraBits`, we must've added a new one at some point a long time ago.
- Reformats `LastLoginAt` to use our standard format (we've had 3 different formats!). Keeps whatever the timestamp was where possible.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,8 +6,42 @@
     "configurations": [
 
 
-
-
+    {
+        "name": "(gdb) pfu",
+        "type": "cppdbg",
+        "request": "launch",
+        "program": "${workspaceFolder}/cmake-build-debug/bin/pfu",
+        "args": [],
+        "stopAtEntry": false,
+        "cwd": "${workspaceFolder}",
+        "environment": [
+            { 
+                "name" : "MUD_DATA_DIR",
+                "value" : "${env:HOME}/tmp/pfiles"
+            },
+            { 
+                "name" : "MUD_AREA_DIR",
+                "value" : "${workspaceFolder}/install/area"
+            },
+            { 
+                "name" : "MUD_HTML_DIR",
+                "value" : "${workspaceFolder}/install/html"
+            },
+            {
+                "name" : "MUD_PORT",
+                "value" : "9000"
+            }
+        ],
+        "externalConsole": false,
+        "MIMode": "gdb",
+        "setupCommands": [
+            {
+                "description": "Enable pretty-printing for gdb",
+                "text": "-enable-pretty-printing",
+                "ignoreFailures": true
+            }
+        ]
+    },
         {
             "name": "(gdb) xania_test",
             "type": "cppdbg",
@@ -45,7 +79,30 @@
                     "ignoreFailures": true
                 }
             ]
-        },        
+        },
+        {
+            "name": "(gdb) apps_test",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/cmake-build-debug/bin/apps_test",
+            "args": [],
+            "stopAtEntry": false,
+            "cwd": "${workspaceFolder}",
+            "environment": [ 
+                { "name" : "MELD_ON_DIFF",
+                  "value" : "1"
+                }
+            ],
+            "externalConsole": false,
+            "MIMode": "gdb",
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                }
+            ]
+        },
         {
             "name": "Debug Xania",
             "type": "cppdbg",
--- a/infra/scripts/prepare-pfiles.sh
+++ b/infra/scripts/prepare-pfiles.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+# This script is run on the mud host as part of a player file upgrade workflow.
+# It creates an archive of the player and gods directories. The archive can be
+# used to restore the original pfiles if the upgrades don't work out.
+# It adds the copied files to a temp git repo then runs `pfu` to perform upgrades on the 
+# player files. You can then review the git diffs before carrying the upgrades
+# back into the production player directory using upgrade-pfiles.sh
+# In dev, run this from the project root directory.
+. "${0%/*}/pfile-upgrade-lib.sh"
+orig_dir=`pwd`
+cd $MUD_DATA_DIR
+if [ -d $snapshot_dir ]; then
+    echo "Snapshot already exists, replacing it."
+    rm -rf $snapshot_dir
+fi
+mkdir -p $snapshot_dir
+echo "Snapshotting player files into $snapshot_dir"
+cp -r player gods $snapshot_dir
+echo "Creating backup tarball of this snapshot: $snapshot_archive_path"
+cd $snapshot_base
+tar Jcf $snapshot_archive_path $snapshot_name
+cd $snapshot_name
+echo "Initializing git repo for player file upgrade"
+git init -q
+git add .
+git commit -qm 'importing pfiles'
+cd $orig_dir
+echo "Running pfu to upgrade the player files in $snapshot_dir"
+# As pfu uses the same configuration as the mud, point MUD_DATA_DIR
+# at the snapshot directory so that it processes the copy of the player files.
+MUD_DATA_DIR=$snapshot_dir pfu
+echo "Review the git diff output in $snapshot_dir then run upgrade-pfiles.sh"
--- a/src/Descriptor.hpp
+++ b/src/Descriptor.hpp
@@ -36,7 +36,6 @@ enum class DescriptorState {
 };
 
 const char *short_name_of(DescriptorState state);
-
 /*
  * Descriptor (channel) structure.
  */
@@ -107,7 +106,9 @@ public:
     void set_endpoint(uint32_t netaddr, uint16_t port, std::string_view raw_full_hostname);
 
     [[nodiscard]] const std::string &host() const noexcept { return masked_host_; }
+    void host(std::string_view host) { masked_host_ = host; }
     [[nodiscard]] std::string login_time() const noexcept;
+    void login_time(Time login_time) { login_time_ = login_time; }
 
     [[nodiscard]] bool flush_output() noexcept;
 
--- a/src/apps/pfu.cpp
+++ b/src/apps/pfu.cpp
@@ -8,6 +8,8 @@
 #include "save.hpp"
 
 #include <dirent.h>
+#include <filesystem>
+#include <iostream>
 #include <magic_enum.hpp>
 #include <optional>
 #include <range/v3/algorithm/fill.hpp>
@@ -16,8 +18,9 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
-
 namespace pfu {
+
+namespace fs = std::filesystem;
 
 // Register new tasks here. Ideally in ascending order of CharVersion. You can have multiple
 // tasks for any given version.
@@ -28,25 +31,20 @@ Tasks register_tasks() {
 }
 
 std::vector<std::string> collect_player_names(const std::string &player_dir, Logger &logger) {
-    DIR *dir = opendir(player_dir.c_str());
-    if (!dir) {
-        logger.critical("Couldn't open player directory: {}", strerror(errno));
+    if (!fs::exists(player_dir) || !fs::is_directory(player_dir)) {
+        logger.critical("Couldn't open player directory: {}", player_dir);
         return {};
     }
     std::vector<std::string> names;
-    struct dirent *entry;
-    struct stat st;
-    while ((entry = readdir(dir))) {
-        if (entry->d_type == DT_REG) {
-            std::string filename = player_dir + entry->d_name;
-            if (!stat(filename.c_str(), &st) && st.st_size > 0) {
-                names.emplace_back(entry->d_name);
+    for (auto &file : fs::directory_iterator(player_dir)) {
+        if (file.is_regular_file()) {
+            if (file.file_size() > 0) {
+                names.emplace_back(file.path().filename().string());
             } else {
-                logger.warn("Ignoring bad player file: {}", filename);
+                logger.warn("Ignoring bad player file: {}", file.path().string());
             }
         }
     }
-    closedir(dir);
     return names;
 }
 
@@ -89,7 +87,6 @@ ResetModifiableAttrs::ResetModifiableAttrs(CharVersion version) : UpgradeTask(ve
 // this was introduced, but it was probably to fix "attribute drift" e.g. where the modifiers of a spell affect
 // got changed in the code without retrospectively adjusting the player's stats.
 void ResetModifiableAttrs::execute(Char &ch) const {
-    OBJ_DATA *obj;
     // Restore the character to his/her true condition
     // All of the attributes reset here are ones that can be modified in AFFECT_DATA.cpp
     ranges::fill(ch.mod_stat, 0);
@@ -104,7 +101,7 @@ void ResetModifiableAttrs::execute(Char &ch) const {
     ch.sex = ch.pcdata->true_sex;
     // add back object effects
     for (auto loc = 0; loc < MAX_WEAR; loc++) {
-        obj = get_eq_char(&ch, loc);
+        auto *obj = get_eq_char(&ch, loc);
         if (!obj)
             continue;
         for (size_t i = 0; i < ch.armor.size(); i++)
@@ -121,8 +118,6 @@ void ResetModifiableAttrs::execute(Char &ch) const {
     for (auto &af : ch.affected)
         af.apply(ch);
 }
-
-CharUpgraderResult::CharUpgraderResult() : ch_(nullptr), upgraded_(false) {}
 
 CharUpgraderResult::CharUpgraderResult(Char *ch, bool upgraded) : ch_(ch), upgraded_(upgraded) {}
 
--- a/src/apps/pfu.cpp
+++ b/src/apps/pfu.cpp
@@ -1,0 +1,195 @@
+#include "pfu.hpp"
+#include "VnumRooms.hpp"
+#include "WrappedFd.hpp"
+#include "common/Configuration.hpp"
+#include "common/Time.hpp"
+#include "handler.hpp"
+#include "merc.h"
+#include "save.hpp"
+
+#include <dirent.h>
+#include <magic_enum.hpp>
+#include <optional>
+#include <range/v3/algorithm/fill.hpp>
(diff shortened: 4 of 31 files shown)


.vscode/launch.json
```diff
@@ -6,8 +6,42 @@
...
+        "type": "cppdbg",
+        "request": "launch",
+        "program": "${workspaceFolder}/cmake-build-debug/bin/pfu",
+        "args": [],
+        "stopAtEntry": false,
+        "cwd": "${workspaceFolder}",
+        "environment": [
+            { 
+                "name" : "MUD_DATA_DIR",
+                "value" : "${env:HOME}/tmp/pfiles"
```
snellers, line 20: Typically when manually running it locally, I'd point it at a copy of the prod pfiles that live here.

src/Descriptor.hpp
```diff
@@ -107,7 +106,9 @@ class Descriptor {
     void set_endpoint(uint32_t netaddr, uint16_t port, std::string_view raw_full_hostname);
 
     [[nodiscard]] const std::string &host() const noexcept { return masked_host_; }
+    void host(std::string_view host) { masked_host_ = host; }
```
snellers, line 109: Added to Descriptor as they're runtime state not tracked in Char itself, and I needed pfu to be able to retain the char's previous login from/time.

src/apps/pfumain.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+extern void boot_db();
+
+// pfu - player file upgrade utility for bulk upgrading all player files
+// as part of a release. Especially where the release contains code changes
+// that will affect character stats or anything else that ought to be
+// persisted in the pfiles.
+// This is intended to be used with the prepare-pfiles.sh wrapper script.
+//
+// There are some limitations right now:
```
snellers, line 18: I've noted some potential things for the future around alternative file formats.

src/apps/test/pfutest.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+struct LoadTinyMudOnce : Catch::TestEventListenerBase {
+    using TestEventListenerBase::TestEventListenerBase;
+
+    void testRunStarting([[maybe_unused]] Catch::TestRunInfo const &testInfo) override {
+        setenv(MUD_AREA_DIR_ENV, TEST_DATA_DIR "/area", 1);
+        setenv(MUD_DATA_DIR_ENV, TEST_DATA_DIR "/data", 1);
+        setenv(MUD_HTML_DIR_ENV, TEST_DATA_DIR "/html", 1);
+        setenv(MUD_PORT_ENV, "9000", 1);
+        boot_db();
```
snellers, line 28: This suite boots up a tiny mud, from `limbo_test.are`. It has just enough rooms, objs and mobs to make this reasonably testable while also being fast.

src/apps/test/pfutest.cpp
```diff
@@ -0,0 +1,133 @@
...
+TEST_CASE("upgrade player") {
+    SECTION("v4 to latest") {
+        const auto all_tasks = register_tasks();
+        const auto player_dir = Configuration::singleton().player_dir();
+        test::MemFile god_file;
+        test::MemFile player_file;
+        const CharSaver saver;
+        auto result =
+            upgrade_player("Versionfour", all_tasks, logger, [&god_file, &player_file, &saver](const Char &ch) {
+                saver.save(ch, god_file.file(), player_file.file());
```
snellers, line 62: Using the templated `upgrade_player()` to inject a saver that takes the output files so we can inspect them. Whereas `pfu` itself calls the single arg version of `save()`.

src/test/fileutils.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
```
snellers, line 1: headerized these functions so I can reuse them easily in the pfu tests

src/apps/pfu.cpp
```diff
@@ -0,0 +1,195 @@
...
+    ch.hitroll = 0;
+    ch.damroll = 0;
+    ch.saving_throw = 0;
+    ch.sex = ch.pcdata->true_sex;
+    // add back object effects
+    for (auto loc = 0; loc < MAX_WEAR; loc++) {
+        obj = get_eq_char(&ch, loc);
+        if (!obj)
+            continue;
+        for (size_t i = 0; i < ch.armor.size(); i++)
```
mattgodbolt, line 110: (not actionable) Darn ... C++ needs a python-like `enumerate()` which lets you get the index too. Elsewhere we use `for (auto &x : armor)` but this needs the index! annoying :)

src/apps/pfu.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+        for (auto &af : obj->affected)
+            af.apply(ch);
+    }
+    // add back spell effects
+    for (auto &af : ch.affected)
+        af.apply(ch);
+}
+
+CharUpgraderResult::CharUpgraderResult() : ch_(nullptr), upgraded_(false) {}
```
mattgodbolt, line 125: This is 100% cool but you could also default `ch_` and `upgraded_` in the header file, then (also in the header file) just declare `CharUpgraderRersult() = default`.

src/apps/pfu.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include "CharVersion.hpp"
+#include <spdlog/spdlog.h>
+#include <vector>
+
+namespace pfu {
+
+class UpgradeTask {
+public:
+    explicit UpgradeTask(const CharVersion upgrade_version) : upgrade_version_(upgrade_version) {}
+    virtual ~UpgradeTask() = default;
```
mattgodbolt, line 12: :+1:

src/apps/pfu.cpp
```diff
@@ -0,0 +1,195 @@
...
+        return {ch, upgraded};
+    }
+    return {};
+}
+
+Char *CharUpgrader::simulate_login() {
+    if (auto fp = WrappedFd::open(filename_for_player(name_))) {
+        LastLoginInfo last_login;
+        // This snippet is very similar to what's in nanny() and try_load_player().
+        Char *ch = new Char();
```
mattgodbolt, line 168: (non-actionable) hopefully soon we can remove these "scary" naked `new`s. But for now we're definitely stuck with them :)

src/apps/pfu.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
```
mattgodbolt, line 1: super nitpick. I put whitespace between pragmas and includes, and separate the blocks so it's 
```cpp
pragma

local includes "..."

library includes <...>

system/STL includes <...>

```

but really that's just my style. The rationale for the ordering (which you've kept to here!) is to spot missing `<STL>` includes in header files, which can be hidden if all the c++ files that include them have already included half the STL before they include the header.

src/apps/pfu.hpp
```diff
@@ -0,0 +1,92 @@
...
+    virtual ~UpgradeTask() = default;
+    [[nodiscard]] bool is_required(const Char &ch) const { return ch.version < upgrade_version_; }
+    virtual void execute(Char &ch) const = 0;
+
+private:
+    const CharVersion upgrade_version_;
+};
+
+// Reset/repair Char attributes that are modifiable via spell/item bonii.
+// This task accepts a version because over time the expectation is that
```
mattgodbolt, line 21: handy comment, thanks.

src/apps/pfu.hpp
```diff
@@ -0,0 +1,92 @@
...
+struct CharUpgraderResult {
+    CharUpgraderResult();
+    CharUpgraderResult(Char *ch, bool upgraded);
+    ~CharUpgraderResult(); // remove the upgraded Char from the world
+    // returns true if the Char was logged in and is valid. false if it couldn't be loaded (e.g. empty file)
+    // and it's unnecessary for any upgrade tasks to have been processed because save should be called for
+    // all valid chars.
+    operator bool() const noexcept;
+    operator Char &() const noexcept;
+    Char *ch_; // set if a Char file was found, but it may still be invalid.
```
mattgodbolt, line 40: as per the comment in the C++ file I'd set these to the defaults here and replec line 32 with `CharUpgradeRestult() = default;`. Just "feels" safer to know whatever goes on, those will be set to sensible defaults. (FUDish sounding, I know)

src/apps/pfu.hpp
```diff
@@ -0,0 +1,92 @@
...
+    bool upgraded_; // true if at least one task was performed
+};
+
+using Tasks = std::vector<std::unique_ptr<UpgradeTask>>;
```
mattgodbolt, line 44: :+1:

src/apps/pfumain.cpp
```diff
@@ -0,0 +1,55 @@
+#include "common/Configuration.hpp"
+#include "handler.hpp"
+#include "pfu.hpp"
+
+#include <fmt/format.h>
+#include <lyra/lyra.hpp>
+#include <spdlog/sinks/stdout_color_sinks.h>
+#include <spdlog/spdlog.h>
+
+extern void boot_db();
```
mattgodbolt, line 10: (non-actionable) one sweet day we'll be able to kill all these kinds of things. we should chat about how to start the process of OOficiation of the MUD. My interim so far has been to introduce (some) singletons, but I'm not sure how to proceed with minimal changes towards a hopeful inevitable "Mud" object comprising all the bits... anyway, one for chat sometime.

src/apps/pfumain.cpp
```diff
@@ -0,0 +1,55 @@
...
+    auto result = cli.parse({argc, argv});
+    if (!result) {
+        fmt::print("Error in command line: {}\n", result.errorMessage());
+        exit(1);
+    } else if (help) {
+        fmt::print("{}", cli);
+        exit(0);
+    }
+    if (verbose) {
+        logger.set_level(spdlog::level::debug);
```
mattgodbolt, line 44: ❤️

src/apps/test/pfutest.cpp
```diff
@@ -0,0 +1,133 @@
+#include "pfu.hpp"
+#include "../test/MemFile.hpp"
```
mattgodbolt, line 2: I'd suggest adding a `target_include_directory` CMake thingy or similar to pick these up. fine as is though, I just "like" not having `../` type relative things in when there's other solutions.

src/apps/test/pfutest.cpp
```diff
@@ -0,0 +1,133 @@
...
+extern void boot_db();
+extern void reap_old_chars();
+
+using namespace pfu;
+
+namespace {
+spdlog::logger logger = spdlog::logger("pfutest");
+}
+
+struct LoadTinyMudOnce : Catch::TestEventListenerBase {
```
mattgodbolt, line 20: Love it! Did you consider using a Fixture instead for this? Or am I missing something? :) Does this run at a different spot than a fixutre? Or is it run only one time and you're trying to avoid taking a long time to run tests?

src/apps/test/pfutest.cpp
mattgodbolt: Got it: I saw the template. I like it :)  I'd have probably "just" tested calling `cup.upgrade()` and then left the actual "save" part untested rather than add machinery to make it tested, but this is probably better :)

src/apps/test/pfutest.cpp
```diff
@@ -0,0 +1,133 @@
...
+                saver.save(ch, god_file.file(), player_file.file());
+            });
+        REQUIRE(result == true);
+        SECTION("upgraded file matches expected") {
+            auto expected = test::read_whole_file(player_dir + "/expected-upgrades/Versionfour");
+            auto actual = player_file.as_string_view();
+
+            CHECK(actual == expected);
+            if (actual != expected)
+                test::diff_mismatch(actual, expected);
```
mattgodbolt, line 71: Nice!!!

src/apps/test/pfutest.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        SECTION("are parsed and converted") {
+            auto parsed = try_parse_login_at(last_login);
+
+            CHECK(parsed == expected);
+        }
+    }
+    SECTION("unsupported timestamp ignored") {
+        std::string last_login = "Feb 22 10:33:21 2000";
+        auto parsed = try_parse_login_at(last_login);
```
mattgodbolt, line 106: I'm so lazy I'd do:
```cpp
REQUIRE(!try_parse_loging("Feb 22 10:33:21 2000");
```
etc for all these. I think each would fit on a single line then :)

src/comm.cpp
```diff
@@ -974,8 +974,6 @@ void nanny(Descriptor *d, const char *argument) {
         d->write("\n\rWelcome to Xania.  May your stay be eventful.\n\r");
         char_list.add_front(ch);
         d->state(DescriptorState::Playing);
-        reset_char(ch);
```
mattgodbolt, line 977: Woo! The meat of the change

src/handler.cpp
```diff
@@ -578,26 +492,25 @@ int apply_ac(OBJ_DATA *obj, int iWear, int type) {
         return 0;
 
     switch (iWear) {
-    case WEAR_BODY: return obj->value[type];
```
mattgodbolt, line 581: oh lol :) Nice

src/save.cpp
```diff
@@ -419,7 +419,10 @@ void fwrite_one_obj(const Char *ch, const OBJ_DATA *obj, FILE *fp, int iNest) {
     fwrite_objs(ch, obj->contains, fp, iNest + 1);
 }
 
-void load_into_char(Char &character, FILE *fp) {
+// Although the LastLogin* fields are read from the player file, it's transient info unused
```
mattgodbolt, line 422: Great comment; thanks.

src/save.cpp
```diff
@@ -660,9 +656,9 @@ void fread_char(Char *ch, FILE *fp) {
         } else if (word == "longdescr" || word == "lnd") {
             ch->long_descr = fread_stdstring(fp);
         } else if (word == "lastloginfrom") {
-            login_from = fread_string(fp);
+            last_login.login_from = fread_stdstring(fp);
```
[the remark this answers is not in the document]
mattgodbolt: So much better, thanks.

src/save.cpp
```diff
@@ -548,12 +544,12 @@ void fread_char(Char *ch, FILE *fp) {
...
                 sn = skill_lookup(fread_word(fp));
                 if (sn < 0)
-                    bug("Fread_char: unknown skill.");
+                    bug("fread_char: unknown skill.");
                 else
                     af.type = sn;
             } else /* old form */
                 af.type = fread_number(fp);
-            if (ch->version == 0)
+            if (ch->version == CharVersion::Zero)
```
mattgodbolt, line 552: Is this even possible now? Seems like not?

src/test/fileutils.hpp
mattgodbolt: 🖖🏻

src/test/fileutils.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <catch2/catch.hpp>
+
+// Some helper functions for dealing with test data files.
+namespace test {
+
+std::string read_whole_file(const std::string &name) {
```
mattgodbolt, line 8: (non-action)
It's frankly embarrassing that in the year of our lord 2021 we have to write this shite, but 100% we do.

src/test/fileutils.hpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// short up deficiencies in diff reporting.
+void diff_mismatch(std::string_view lhs, std::string_view rhs) {
+    // If you find yourself debugging, set this ENV var to run meld.
+    if (!::getenv("MELD_ON_DIFF"))
+        return;
+    write_file("/tmp/lhs", lhs);
+    write_file("/tmp/rhs", rhs);
+    CHECK(system("meld /tmp/lhs /tmp/rhs") == 0);
```
mattgodbolt, line 39: I love this. c.f. [Approval Tests](https://approvaltests.com/) (which one day might be useful, but not yet)

src/test/save_test.cpp
```diff
@@ -2,51 +2,14 @@
...
 #include "merc.h"
 #include "string_utils.hpp"
 
 #include <catch2/catch.hpp>
 
 using namespace std::literals;
 
-namespace {
-
-std::string read_whole_file(const std::string &name) {
```
mattgodbolt, line 14: Ah I see this moved! :blush:

mattgodbolt (approved):
Looks great! 33 comments; sorry. Most of those I think are positive though! A couple super minor things in there though :) Great stuff!

infra/scripts/prepare-pfiles.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# This script is run on the mud host as part of a player file upgrade workflow.
+# It creates an archive of the player and gods directories. The archive can be
+# used to restore the original pfiles if the upgrades don't work out.
+# It adds the copied files to a temp git repo then runs `pfu` to perform upgrades on the 
+# player files. You can then review the git diffs before carrying the upgrades
+# back into the production player directory using upgrade-pfiles.sh
+# In dev, run this from the project root directory.
+. "${0%/*}/pfile-upgrade-lib.sh"
```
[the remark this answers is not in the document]
snellers: I did test it with the infra/scripts dir in my PATH (relative and absolute versions) and it seemed to work ok. That said, I'd be running it by specifying the path to it anyway.

src/save.cpp
snellers: Yeah I think so, killing

src/apps/test/pfutest.cpp
snellers: Yeah, using the event listener with `testRunStarting`, it only runs once. Whereas from what I can see, specifying a test fixture class as part of TEST_CASE will create an instance of it for each case/method you specify it on. Not quite what I was looking for.  Added to that, due to the current global nature of the mud world, it'll barf if you try to load the same mud entities more than once. Fixable if we have a proper lifecycle for the mud at some point.

PR #264, opened 2021-09-27
Use gcc 10. Drop Ubuntu 18 as we no longer use it anywhere.

snellers:


Makefile
```diff
@@ -22,6 +22,8 @@ CONAN_VERSION=1.38.0
 PIP:=$(CONDA_ROOT)/bin/pip
 CONAN:=$(CONDA_ROOT)/bin/conan
 SOURCE_FILES:=$(shell find src -type f -name \*.c -o -name \*.h -o -name \*.cpp -o -name *.hpp)
+export CC:=gcc-10
```
snellers, line 25: Given that the build is combination of make & cmake rather than just cmake by itself, specifying the compiler this way seems to work. Although I've seen you can do it one or two other ways.

mattgodbolt: Works for me, though I'll probably have to override it locally if/when we get to gcc 11 (but I'll worry about that)

mattgodbolt (approved):

Makefile
snellers: Yep, sounds good.

PR #265, opened 2021-09-28
Extract a Char try_give command

### LINKED ISSUE #262: Make do_give() use ArgParser

mattgodbolt:
* Implement do_give with ArgParser (closes #262)
* rephrases magic to use new function instead of building a string.

### DIFF
--- a/src/Char.cpp
+++ b/src/Char.cpp
@@ -25,6 +25,8 @@
 #include "TimeInfoData.hpp"
 #include "WearLocation.hpp"
 #include "act_comm.hpp"
+#include "act_obj.hpp"
+#include "act_wiz.hpp"
 #include "comm.hpp"
 #include "common/BitOps.hpp"
 #include "db.h"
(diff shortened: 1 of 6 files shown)


src/Char.cpp
```diff
@@ -512,4 +514,47 @@ bool Char::is_starving() const noexcept { return is_pc() && pcdata->hunger.is_un
...
+    }
+
+    if (!can_see_obj(victim, object)) {
+        ::act("$N can't see it.", this, nullptr, victim, To::Char);
+        return;
+    }
+
+    obj_from_char(object);
+    obj_to_char(object, victim);
+    MOBtrigger = false;
```
snellers, line 552: The MOBtrigger global eventually needs to be nuked :(

src/act_obj.hpp
```diff
@@ -16,3 +16,4 @@ bool obj_move_violates_uniqueness(Char *source_char, Char *dest_char, Object *mo
 bool obj_move_violates_uniqueness(Char *source_char, Char *dest_char, Object *moving_obj,
                                   GenericList<Object *> &objs_to);
 bool check_material_vulnerability(Char *ch, Object *object);
+void handle_corpse_summoner(Char *ch, Char *victim, Object *obj);
```
snellers, line 19: This seems fine although it crossed my mind that we could also move it to CorpseSummoner.hpp if only for consistency (I'm happy to look at that separately).

snellers (approved):

PR #269, opened 2021-10-19
Refactor the AREA_DATA

mattgodbolt:
* Make it into a class
* Give it some behaviour


mattgodbolt:
Thanks chap! I had a few more things so I've done those here. Will merge later if no probs

PR #273, opened 2021-10-31
Rephrase pers()

mattgodbolt:
Add tests

### DIFF
--- a/src/Char.cpp
+++ b/src/Char.cpp
@@ -559,3 +559,9 @@ void Char::try_give_item_to(Object *object, Char *victim) {
     handle_corpse_summoner(this, victim, object);
     mprog_give_trigger(victim, this, object);
 }
+
+std::string_view Char::describe(const Char &to_describe) const noexcept {
+    if (!can_see(to_describe))
+        return "someone"sv;
+    return to_describe.is_npc() ? to_describe.short_descr : to_describe.name;
+}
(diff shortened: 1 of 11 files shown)


src/act_info.cpp
```diff
@@ -230,7 +230,9 @@ void show_char_to_char_0(const Char *victim, const Char *ch) {
         else if (victim->fighting == ch)
             buf += " |RYOU!|w";
         else if (victim->in_room == victim->fighting->in_room) {
-            buf += fmt::format(" {}.", pers(victim->fighting, ch));
+            std::string_view result;
+            result = ch->describe(*victim->fighting);
```
snellers, line 234: nitpick: the assignment could be one line up.

src/Char.cpp
```diff
@@ -559,3 +559,9 @@ void Char::try_give_item_to(Object *object, Char *victim) {
     handle_corpse_summoner(this, victim, object);
     mprog_give_trigger(victim, this, object);
 }
+
+std::string_view Char::describe(const Char &to_describe) const noexcept {
```
snellers, line 563: much nicer!

snellers (approved):

src/act_info.cpp
mattgodbolt: not a nitpick! that's the automatic tooling going wrong... there shouldn't be an intermediate variable at all.

src/Char.cpp
mattgodbolt: Can be the basis of a lot of things that can be described too, like ...describe(Room &room) etc.

Also I wondered about trying to find a way to "just" "know" which Char is active in e.g. `send_to()` etc, so that ideally we could just do `send_line("{} is here.", other);` and have the formatter "know" that it's formatting for "*this" etc. but I couldn't see how to do that without an evil global.

PR #276, opened 2021-11-12
Introduce a validator for checking scroll spell targets along with a

snellers:
general way of adding validators that are run when loading areas. Fix
a few invalid scrolls.

### DIFF
--- a/area/HighTowerOfSorcery.are
+++ b/area/HighTowerOfSorcery.are
@@ -4820,7 +4820,7 @@ a scroll written on blue paper~
 A blue scroll has been left here.~
 vellum~
 scroll G A
-35 24 1 1 0
+35 24 0 0 0
 23 5 4678 P
 
 #1364
--- a/src/ObjectIndexValidator.hpp
+++ b/src/ObjectIndexValidator.hpp
@@ -1,0 +1,27 @@
+/************************************************************************/
+/*  Xania (M)ulti(U)ser(D)ungeon server source code                     */
+/*  (C) 2021 Xania Development Team                                     */
+/*  See the header to file: merc.h for original code copyrights         */
+/************************************************************************/
+#pragma once
+
+#include <optional>
+#include <string>
+
+class ObjectIndex;
+// An interface for validators that can scrutinise an ObjectIndex after it's
+// been loaded from an area file.
+class ObjectIndexValidator {
+public:
+    virtual ~ObjectIndexValidator() = default;
+
+    // The result's error_message is populated if validation failed.
+    struct Result {
+        std::optional<std::string> error_message;
+    };
+
+    [[nodiscard]] virtual Result validate(const ObjectIndex *obj_index) const = 0;
+
+protected:
+    const Result Success = ObjectIndexValidator::Result{std::nullopt};
+};
--- a/src/ScrollTargetValidator.cpp
+++ b/src/ScrollTargetValidator.cpp
@@ -1,0 +1,76 @@
+/************************************************************************/
+/*  Xania (M)ulti(U)ser(D)ungeon server source code                     */
+/*  (C) 2021 Xania Development Team                                     */
+/*  See the header to file: merc.h for original code copyrights         */
+/************************************************************************/
+#include "ScrollTargetValidator.hpp"
+#include "ObjectIndex.hpp"
+#include "ObjectType.hpp"
+#include "Target.hpp"
+
+#include <fmt/core.h>
+
+ScrollTargetValidator::ScrollTargetValidator(const struct skill_type *skill_table)
+    : skill_table_(skill_table),
+      incompatible_targets_{
+          // clang-format off
+          {Target::ObjectInventory, Target::CharOffensive},
+          {Target::ObjectInventory, Target::CharSelf},
+          {Target::ObjectInventory, Target::CharOther},
+          {Target::ObjectInventory, Target::CharDefensive},
+          {Target::ObjectInventory, Target::CharObject},
+          {Target::CharObject, Target::CharOffensive},
+          {Target::CharObject, Target::CharSelf},
+          {Target::CharObject, Target::CharOther},
+          {Target::CharObject, Target::CharDefensive},
+          // clang-format on
+      } {}
+
+ObjectIndexValidator::Result ScrollTargetValidator::validate(const ObjectIndex *obj_index) const {
+    if (obj_index->type != ObjectType::Scroll) {
+        return Success;
+    }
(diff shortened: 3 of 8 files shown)


area/HighTowerOfSorcery.are
```diff
@@ -4820,7 +4820,7 @@ a scroll written on blue paper~
 A blue scroll has been left here.~
 vellum~
 scroll G A
-35 24 1 1 0
+35 24 0 0 0
```
snellers, line 4823: We had a few scrolls of enchant weapon paired with armour.

mattgodbolt: Nice spot!

src/ObjectIndexValidator.hpp
```diff
@@ -0,0 +1,27 @@
...
+class ObjectIndexValidator {
+public:
+    virtual ~ObjectIndexValidator() = default;
+
+    // The result's error_message is populated if validation failed.
+    struct Result {
+        std::optional<std::string> error_message;
+    };
+
+    [[nodiscard]] virtual Result validate(const ObjectIndex *obj_index) const = 0;
```
mattgodbolt, line 23: Nice! I love the named `error_message` to make it super clear what the result is.

src/ScrollTargetValidator.cpp
```diff
@@ -0,0 +1,76 @@
...
+/*  See the header to file: merc.h for original code copyrights         */
+/************************************************************************/
+#include "ScrollTargetValidator.hpp"
+#include "ObjectIndex.hpp"
+#include "ObjectType.hpp"
+#include "Target.hpp"
+
+#include <fmt/core.h>
+
+ScrollTargetValidator::ScrollTargetValidator(const struct skill_type *skill_table)
```
mattgodbolt, line 13: I'd drop the unnecessary `struct` here, and for new code I've started using references where I can instead of the super-prevalent `pointer but I'm going to assume it's not null` that the rest of the code base uses.

```suggestion
ScrollTargetValidator::ScrollTargetValidator(const skill_type &skill_table)
```

etc :) Seriously no biggie, this is great as is. There's a few things here that could take a `const &` too, but won't ping them all. And I'm sure I've been inconsistent in my new code too...so...a suggestion at best :)

src/ScrollTargetValidator.cpp
```diff
@@ -0,0 +1,76 @@
...
+    if (are_incompatible(opt_target2, opt_target3)) {
+        return ObjectIndexValidator::Result{
+            fmt::format("Incompatible target type for slot2 and slot3 in #{}", obj_index->vnum)};
+    }
+    return Success;
+}
+
+bool ScrollTargetValidator::are_incompatible(const std::optional<Target> &opt_target_a,
+                                             const std::optional<Target> &opt_target_b) const {
+    const auto are_mapped = [this](const auto key, const auto value) {
```
mattgodbolt, line 54: the `const auto` here will take a copy of `key` and `value`; you can probably get away with a `const auto &` in both cases here.

src/ScrollTargetValidator.cpp
```diff
@@ -0,0 +1,76 @@
...
+            if (value == it->second)
+                return true;
+        }
+        return false;
+    };
+
+    if (!opt_target_a || !opt_target_b) {
+        return false;
+    }
+    const auto target_a = *opt_target_a;
```
mattgodbolt, line 66: similarly here you're copying the `Target` object here. Which might be find (in honestly I haven't looked down the code yet at what a `Target` is), but you could do `const auto &target_a = &opt_ptarget_a;` etc here.

src/ScrollTargetValidator.hpp
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include "ObjectIndexValidator.hpp"
+#include "SkillTables.hpp"
+
+#include <map>
+#include <optional>
+
+class ObjectIndex;
+enum class Target;
```
mattgodbolt, line 15: Ah! it's `Target` is only an `enum` ... very much weakens my argument above. Sorry (leaving the comments there for discussion only)

src/ScrollTargetValidator.cpp
```diff
@@ -0,0 +1,76 @@
...
+    const auto target_b = *opt_target_b;
+    return are_mapped(target_a, target_b) || are_mapped(target_b, target_a);
+}
+
+std::optional<Target> ScrollTargetValidator::get_target(const ObjectIndex *obj_index, const int index) const {
+    if (obj_index->value[index] >= 0) {
+        return skill_table_[obj_index->value[index]].target;
+    }
+    return std::nullopt;
+}
```
mattgodbolt, line 76: nit, missing `\n`

src/ScrollTargetValidator.cpp
mattgodbolt: (update post first reading...the `Target` is probably just fine to copy around)

src/db.cpp
```diff
@@ -710,7 +711,8 @@ void load_objects(FILE *fp) {
         bug("Load_objects: no #AREA section found yet!");
         exit(1);
     }
-
+    const auto validators = std::array<std::unique_ptr<const ObjectIndexValidator>, 1>{
```
mattgodbolt, line 714: Nice

src/test/ScrollTargetValidatorTest.cpp
```diff
@@ -0,0 +1,213 @@
...
+    SECTION("valid not a scroll") {
+        obj_index.type = ObjectType::Wand;
+        obj_index.value[1] = FireballIndex;
+        obj_index.value[3] = EnchantIndex;
+
+        const auto result = validator.validate(&obj_index);
+
+        CHECK(!result.error_message);
+    }
+}
```
mattgodbolt, line 213: nit: missing newline

src/test/ScrollTargetValidatorTest.cpp
```diff
@@ -0,0 +1,213 @@
...
+namespace {
+
+const auto EmptyIndex = -1;
+const auto FireballIndex = 0;
+const auto EnchantIndex = 1;
+const auto ForgeIndex = 2;
+const auto FaerieFogIndex = 3;
+const auto DetectEvilIndex = 4;
+
+struct skill_type test_skill_table[5] = {
```
mattgodbolt, line 23: unnecessary `struct` and maybe use `std::array` ? (though that makes constructing the validator need some gymnastics, so maybe not?)
```suggestion
const std::array<skill_type, 5> test_skill_table {
```

mattgodbolt (approved):
Looks great

PR #282, opened 2021-12-10
Officially switch to C++20

frederic-tingaud-sonarsource:
The project is already using C++20 features. It's just that GCC10 is a little too lenient on that.

That switch adds a few more constraints on fmt.


frederic-tingaud-sonarsource:
The C++20 pattern that was already there:
https://godbolt.org/z/6sWv1K7e3

snellers:
> 

Nice!

PR #283, opened 2021-12-11
Fix compilation with g++11

OznOg:


snellers:
@OznOg LGTM. I've been able to verify on Ubuntu 20 with g++11 too although 11 is not quite "official" for the mud yet, and isn't enabled in the CI yet either. But I think it can be quite soon.

snellers (approved):

PR #287, opened 2021-12-18
Some cmake rework

OznOg:
Remove bad use of include_directories and enforce dependencies visibility.


src/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
-configure_file(version.h.in version.h @ONLY)
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
+configure_file(version.h.in generated_sources/version.h @ONLY)
+
+add_library(version INTERFACE ${CMAKE_CURRENT_BINARY_DIR}/generated_sources/version.h)
```
snellers, line 3: `add_library` of type `INTERFACE` with sources specified didn't appear in cmake until v3.19.  In our CI it always installs the latest cmake so it works there. On my box I was still using 3.18 so it barfed. I will be updating the README with an extra step about how to grab the latest (3.22+) from the kitware apt repo.

src/chat/test/DatabaseTest.cpp
```diff
@@ -5,8 +5,8 @@
 /*  Chat bot originally written by Chris Busch in 1993-5, this file is a */
 /*  reimplementation of that work.                                       */
 /*************************************************************************/
-#include "Database.hpp"
-#include "KeywordResponses.hpp"
+#include "chat/Database.hpp"
```
snellers, line 8: I like that the headers use qualified paths now :)

snellers:
LGTM and verified locally.

src/CMakeLists.txt
OznOg: That's a good point, it would also be possible to workaround the source in interfact by using target_sources or else. Let me know if you want this to be done.

PR #290, opened 2021-12-23
Draft: Simplify build

OznOg:
This PR should allow to compile xania with or without Conan, allowing to create correct packaging in distros.
It separates the "setup" part (pull dependencies and provide them to cmake) form pure compilation and libraries detection.
One may also use conan to setup a build env and use only cmake later on; this prevents useless conan chacks on every make call.


OznOg:
This PR should allow to compile xania with or without Conan, allowing to create correct packaging in distros.
It separates the "setup" part (pull dependencies and provide them to cmake) form pure compilation and libraries detection.
One may also use conan to setup a build env and use only cmake later on; this prevents useless conan chacks on every make call.

CMakeLists.txt
```diff
@@ -1,56 +1,34 @@
...
 
 set(BUILD_VERSION "(local build)" CACHE STRING "Build version (build number only)")
 set(BUILD_FULL_VERSION "(local build)" CACHE STRING "Full string of build version (with git info)")
 set(BUILD_COMMIT_URL "https://this.is.a.local.build/1234" CACHE STRING "Full string of git URL to this build")
 
+project(xania)
+
 set(CMAKE_CXX_STANDARD 20)
-add_compile_options(-Werror -Wall -Wextra)
+add_compile_options(-Werror -Wall -Wextra -Wno-error=deprecated-declarations)
```
snellers, line 10: We're not using `[[deprecated]]` in any of the mud sources,  are you getting errors related to this from a library or something?

CMakeLists.txt
```diff
@@ -1,56 +1,34 @@
 cmake_minimum_required(VERSION 3.19)
-project(xania)
-
-####
-# Shell out to the Makefile to get all the paths it installed for us
-set(CMAKE_MAKE_CONFIG_FILE ${CMAKE_CURRENT_BINARY_DIR}/config-from-make.cmake)
```
snellers, line 6: I checked out the `pr/290` branch on my box and did a distclean, then `make install test` -- broken in a couple of ways currently. First, seems to report using gcc/g++-9, even though I have 10 and 11 installed:

```
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0

```

Second, it barfs on conan:
```
-- Conan: checking conan executable
-- Conan: Found program /home/rsnell/.local/bin/conan
CMake Error at conan.cmake:807 (message):
  Conan --version failed='1'
Call Stack (most recent call first):
  CMakeLists.txt:13 (conan_check)

```

snellers: OK I figured out why I was getting that conan.cmake error -- I had an ancient version of the `conan` wrapper script left over in my `~/.local/bin/` directory, I think from when I had attempted to install conan directly myself using `pip3` last year.  I've removed this and that error goes away, and it attempts to do the conan download instead.  However that is failing with permission denied like this:

```
'chmod' '+x' '/home/rsnell/dev/xania/cmake-build-debug/conda-4.10.3-installer.sh'
'/home/rsnell/dev/xania/cmake-build-debug/conda-4.10.3-installer.sh' '-u' '-b' '-p' '/home/rsnell/dev/xania/cmake-build-debug/.tools//conda-4.10.3'
CMake Error at CMakeLists.txt:36 (execute_process):
  execute_process error getting child return code: Permission denied

```

snellers: Looks like it's failing in the CI for the same reason.

CMakeLists.txt
```diff
@@ -1,56 +1,34 @@
...
-
-include("${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
-conan_basic_setup(TARGETS)
-
-set(CMAKE_INSTALL_MESSAGE NEVER) # prevent logging installation messages
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=leak -fsanitize=address -fsanitize=undefined -static-libasan ")
 
 enable_testing()
 
+find_package(spdlog REQUIRED)
```
snellers, line 20: How can we tell it to use a specific version of these libs?  That's something we could do with today's build.

snellers: Oh, I see, conanfile.txt is still there. I guess it means that for anyone not relying on conan to install those 3rd party packages, they might end up using an older/wrong version of one of the packages?

CMakeLists.txt
OznOg: yes I've had it with fmt upto date. That said, I will probably remove this from the PR in it final form as it have nothing to do in  it (compile flags can be overriden at configuration)

snellers: ok

CMakeLists.txt
OznOg: yes, I'm currently trying to fix that, but I'm testing with the network from my phone and the conda dowloads is difficult to test because of the (very) limited bandwidth... I think I'll need to wait I'm back home for testing correctly the network parts...

snellers: No worries, there is no urgency :)

CMakeLists.txt
OznOg: actually, you don't do it in cmake but in conan. That said, this is normally feasible in cmake, and would probably be a good idea. To be continued :)

CMakeLists.txt
OznOg: FYI the warning is issued by lyra (using the current master)
xania/src/doorman/main.cpp:33:70: warning: 'std::string lyra::detail::result_base::errorMessage() const' is deprecated [-Wdeprecated-declarations]
   33 |         fmt::print("Error in command line: {}\n", result.errorMessage());
      |                                                   ~~~~~~~~~~~~~~~~~~~^~
In file included from /tmp/lyra_install/include/lyra/parser_result.hp
...
                 from /tmp/lyra_install/include/lyra/lyra.hpp:12,
                 from /home/marie/trucaseb/git/xania/src/doorman/main.cpp:21:
/tmp/lyra_install/include/lyra/detail/result.hpp:22:36: note: declared here
   22 |         [[deprecated]] std::string errorMessage() const { return message(); }
Not sure anything is needed on xania's side, I give it for the record...

CMakeLists.txt
OznOg: should be fixed now

snellers (approved):

PR #305, opened 2023-08-06
object_list now owns all Objects. Add Object constructor.

snellers:


src/db.cpp
```diff
@@ -1191,88 +1190,17 @@ void clone_mobile(Char *parent, Char *clone) {
...
-    case ObjectType::Potion:
-    case ObjectType::Pill:
-    case ObjectType::Money: break;
-    }
-
-    object_list.add_front(obj);
-    objIndex->count++;
-
-    return obj;
+    auto obj = std::make_unique<Object>(objIndex);
```
snellers, line 1200: @mattgodbolt  I was keen to eliminate more naked news/deletes. I've done it for Object like so, and tweaked GenericList so that the global object_list now owns all of the Object pointers. Sorted out a constructor too. I hope you're ok with this kind of approach.

src/save.cpp
```diff
@@ -581,9 +581,7 @@ void fread_obj(Char *ch, FILE *fp, ObjectNestMap &nest_level_to_obj) {
     bool fNest = false;
     bool fVnum = false;
     bool first = true; /* used to counter fp offset */
-    bool new_format = false; /* to prevent errors */
```
snellers, line 584: new_format hasn't been needed for ages

